PR #26, opened 2019-01-14
WIP: Travis docker

justusschock:


justusschock:
Maybe we should move tests outside docker, because currently all tests in docker would work on master branch which does not make sense for other branches as this would not cover changes made on these branches. What do you think @haarburger 

EDIT: Another way to go, would be to mount the current directory (which includes the local branch of the repo) and build the image with this (this would not make the dockerfile as standalone work)

PR #34, opened 2019-01-16
Kfold + dataset update

justusschock:
Add methods to extract subsets from dataset and datamanager and update kfold accordingly.

### DIFF
--- a/delira/training/experiment.py
+++ b/delira/training/experiment.py
@@ -183,6 +183,41 @@ class AbstractExperiment(TrixiExperiment):
                          data: BaseDataManager,
                          num_splits=None, shuffle=False, random_seed=None,
                          label_key="label", **kwargs):
+         """
+        Runs stratified K-Fold Crossvalidation
+        The supported supported scenario is:
+
+            * passing a single datamanager: the data within the single manager
+              will be split and multiple datamanagers will be created holding
+              the subsets.
+
+        .. note:: 
+            In opposite to :method:`AbstractExperiment.kfold` this method does 
+            not support passing multiple managers, since stratification would 
+            be hard to implement
+
+        Parameters
+        ----------
+        num_epochs : int
+            number of epochs to train the model
+        data : :class:`BaseDataManager`
+            list of datamanagers or single datamanger
+            (will be split for crossvalidation)
+        num_splits : None or int
+            number of splits for kfold
+            if None: len(data) splits will be validated
+        shuffle : bool
+            whether or not to shuffle indices for kfold
+        random_seed : None or int
+            random seed used to seed the kfold (if shuffle is true),
+            pytorch and numpy
+        label_key : str (default: "label")
+            the key to extract the label for stratification from each data 
+            sample
+        **kwargs :
+            additional keyword arguments (completely passed to self.run())
+
+        """
 
         if num_splits is None:
             num_splits = 10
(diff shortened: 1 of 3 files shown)


delira/training/experiment.py
```diff
@@ -94,17 +94,31 @@ class initialization
...
+        The two supported scenarios are:
+
+            * passing a list of datamanagers: the list will be splitted and 
+            multiple datamanagers will be combinded into a single one 
+            (if necessary); No subsets will be created from a datamanager but 
+            each train and validation set consists of a integer number of full 
+            datamanagers
+
+            * passing a single datamanager: the data within the single manager 
+            will be splitted and multiple datamanagers will be created holding
```
ORippler: typo: will be split

delira/training/experiment.py
```diff
@@ -94,17 +94,31 @@ class initialization
...
 
         Parameters
         ----------
         num_epochs : int
             number of epochs to train the model
-        data : list of BaseDataManager
-            list of datamanagers (will be split for crossvalidation)
+        data : list of :class:`BaseDataManager` or single 
+               :class:`BaseDataManager`
+            list of datamanagers or single datamanger
```
ORippler: typo datamanger

delira/training/experiment.py
```diff
@@ -118,20 +132,48 @@ def kfold(self, num_epochs: int, data: typing.List[BaseDataManager],
 
         """
 
+        # set number of splits if not specified
         if num_splits is None:
-            num_splits = len(data)
-
+            if isinstance(data, BaseDataManager):
+                num_splits = len(data.dataset)
```
ORippler: I don't think setting num_splits = len(dataset) is a good idea. If for example we have 5000 training samples we will get 5000 splits. Better throw a warning and reduce it to some commonly used number like 5 or at most 10.

ORippler (changes requested):
In general: How easy is it to extract the labels from `datamanger.dataset` ? If so, one could think of also adding `StratifiedKfold` to ensure equal label distributions amongst splits. I think with `BaseLazyDataset` this could probe to be difficult though.

justusschock:
To add a `StratifiedKfold` we would have to. Iterate over all data samples once. While this could take a while, it should not be difficult to implement.

We could raise a Warning if the dataset is a lazy one, but this overhead must be made only once, I think we should add it.

delira/training/experiment.py
justusschock: Fixed

delira/training/experiment.py
justusschock: fixed

delira/training/experiment.py
justusschock: Changed default value to 10 splits

justusschock:
cc @ORippler . Requested changes have been made

delira/training/experiment.py
```diff
@@ -177,6 +179,74 @@ def kfold(self, num_epochs: int,
...
+        Parameters
+        ----------
+        num_epochs : int
+            number of epochs to train the model
+        data : :class:`BaseDataManager`
+            single datamanager
+            (will be split for crossvalidation)
+        num_splits : None or int
+            number of splits for kfold
+            if None: len(data) splits will be validated
```
ORippler: if None: 10 splits

ORippler (changes requested):
update docstring.

also [here](https://github.com/justusschock/delira/blob/e5dc2e79aa9281b5d9d931b62f5430f7ec01933a/delira/training/experiment.py#L124)

delira/training/experiment.py
justusschock: done

ORippler (approved):
Seems good to me now

Issue #35, opened 2019-01-18
[FeatureRequest] Define seperate transformations for train and test data in k-fold.


mibaumgartner:
The k-fold method in the experiment class currently only works with datamanagers with pre-defined transformations. 
It would be great, if you could define some base-transformations (e.g. pre-processing) which are applied to the train and test set and train-transformations (e.g. augmentation) which are only applied to the training set.

https://github.com/justusschock/delira/blob/3374ac6b86358b6910f3ac1a8c070888e8dc9c33/delira/training/experiment.py#L97-L113

PR #37, opened 2019-01-18
Add delira config 

justusschock:
Add configuration file and parser in `delira.__init__` 
Make dependency check based on environment variable


justusschock:
I have not found a way to generate a file at installation time and thus it will be checked (and created if necessary) at import time

EDIT: change basic requirements to `trixi-slim` and add `trixi` to torch requirements (https://github.com/MIC-DKFZ/trixi/issues/7)

delira/data_loading/data_manager.py
```diff
@@ -245,108 +244,111 @@ def n_batches(self):
...
-        self.batch_size = datamanager[0].batch_size
-        self.n_process_augmentation = datamanager[0].n_process_augmentation
-        self.transforms = datamanager[0].transforms
-        self.sampler = datamanager[0].sampler.__class__.from_dataset(
-            self.dataset
-        )
-
-    def get_batchgen(self, seed=1):
+    from torch.utils.data import ConcatDataset
+    class ConcatDataManager(object):
```
ORippler: in what way does ConcatDataManager depend on `pytorch`?

ORippler (commented):
If `ConcatDataset` is from `torch`, then we must refactor the [`AbstractExperiment`](https://github.com/justusschock/delira/blob/5ae2bd9318caa1822efb164327c8b16e1301e856/delira/training/experiment.py#L26), as it heavily utilizes above class

delira/data_loading/data_manager.py
justusschock: The used ConcatDataset is currently taken from pytorch. We might want to change that in future.

But it should be easy to provide a custom implementation.

ORippler: Looking at the pytorch docs, [ConcatDataset](https://pytorch.org/docs/stable/_modules/torch/utils/data/dataset.html#ConcatDataset) itself also doesn't depend on any torch-specific functions. Imo we should add a custom implementation.

justusschock: I'll do that today or tomorrow

justusschock: Should be fixed by 8e364e5b374a0656bce307fe445931b6a3057e8c

delira/data_loading/dataset.py
```diff
@@ -442,6 +463,52 @@ def __getitem__(self, index):
...
+            sample corresponding to given index
+        """
+
+        curr_max_index = 0
+        for dset in self.data:
+            prev_max_index = curr_max_index
+            curr_max_index += len(dset)
+
+            if prev_max_index <= index < curr_max_index:
+                return dset[index % curr_max_index]
```
ORippler: should be `dset[index % prev_max_index]`

delira/data_loading/data_manager.py
```diff
@@ -245,108 +245,81 @@ def n_batches(self):
             raise ValueError('Invalid value for n_process_augmentation')
         return n_batches
 
-
-class ConcatDataManager(object):
+class ConcatDataManager(BaseDataManager):
```
ORippler, line 251: I don't quite get the use-case for the `ConcatDataManager`. The current implementation just concatenates all datasets with a `ConcatDataset` and then then merges them under the same `BaseDataManager`. For me, conceptually, `ConcatDataManager` should aggregate the individual `BaseDataManager` while retaining their initial configuration.

I.e. I have 2 Datasets (MRI + CT), and want to apply different transforms to each of them, therefore I setup 2 different `BaseDataManager` and combine them with `ConcatDataManager`, and use the result for training

delira/data_loading/dataset.py
ORippler: fixed by 836f838e2ad32bcf6702903bcebacd780657f454

delira/data_loading/data_manager.py
justusschock: Currently the usecase is to concatenate datasets which are already wrapped into `DataManager`s. 
IMO it does not make sense to keep the single managers and aggregate them, because there is no indexing performed and we would have to use additional samplers.
Additionally, it is (AFAIK) impossible to apply different transforms to subsets of the datamanager because the transforms are part of the `MultithreadedAugmenter` of `batchgenerators`. The transforms you talk must be applied inside the dataset, which are not the same transforms as the ones of the datamanager, which are mainly for augmentation purposes.

ORippler: Okay, then I would argue that the cleaner solution is to aggregate the datasets once with `ConcatDataset` and instantiate a new `BaseDataManager` with the `ConcatDataset` and other desired parameters and remove the `ConcatDataManager` in a future release

ORippler (approved):
Looks good now, should remove `ConcatDataManager` and support for it when implementing #35

Issue #17, opened 2019-01-08
Hyperparameter search


justusschock:
Integrate hyperparameter tuning with ray

PR #42, opened 2019-01-22
Added test function to experiment

mibaumgartner:
CC: @justusschock @ORippler

### DIFF
--- a/delira/training/experiment.py
+++ b/delira/training/experiment.py
@@ -530,7 +530,7 @@ try:
         def test(self,
                  params: Parameters,
                  network: AbstractPyTorchNetwork,
-                 datamgr_test: typing.Union[BaseDataManager, ConcatDataManager],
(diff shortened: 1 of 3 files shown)


justusschock (commented):
I don't see any problems right now. We might have to refactor this, once #17 is finished, but the changes should be minor, so if @ORippler agrees, this could be a workaround to go.

delira/training/experiment.py
```diff
@@ -527,6 +527,71 @@ def run(self,
                                                  "lowest")
                                  )
 
+        def test(self,
```
justusschock, line 571: Add abstract method in `AbstractExperiment` to ensure consistent API

justusschock (changes requested):

justusschock (approved):

ORippler (approved):

Issue #47, opened 2019-01-29
[FeatureRequest] Add further features to Tf backend


ORippler:
**Description**
Some nice features that are still missing:
* [ ]  Multi-GPU training
* [ ]  Base Segmentation Network
* [ ]  Base GAN Network
* [ ] [Pretrained Models/Networks](https://github.com/tensorflow/models/tree/master/official), must be implemented by means of `tf.keras.Model`, as that is where the long term support is heading for tf
* [x] Logging frequency during training

ORippler:
logging frequency was duplicated and integrated #48 
multi gpu has its own FR #177  and the other features belong to delira_vision

Issue #45, opened 2019-01-24
Tf Integration


justusschock:
Currently working on this

justusschock:
Currently todo:

- [x] Rebase from master and refactor accordingly (e.g. #6)
- [x] Add docstrings similar to the existing ones ([Numpy style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html))
- [x] Add tests with pytest (and ensure tests are skipped if not enough RAM etc.); examples can be found at the [currently implemented tests](https://github.com/justusschock/delira/tree/master/tests)
- [x] Add a requirements file for all tensorflow requirements similar to the [torch requirements file](https://github.com/justusschock/delira/blob/master/requirements_extra_torch.txt)
- [x] Register it as backend in [`setup.py`](https://github.com/justusschock/delira/blob/master/setup.py) and [`__init__.py`](https://github.com/justusschock/delira/blob/master/delira/__init__.py)
- [ ] maybe: Combine classes with same purpose but different backend in one file for simplicity (esp. in [models](https://github.com/justusschock/delira/blob/master/delira/models) and [training](https://github.com/justusschock/delira/blob/master/delira/training))

ORippler:
tests should pass once [tensorboard logging is released in trixi](https://github.com/MIC-DKFZ/trixi/issues/12)

PR #49, opened 2019-01-29
Tf integration

ORippler:
Should add a basic implementation of tf backend to delira, solving #45 and adding:

* `IO`
* `TfNetwork` classes (both Abstract and for Classification)
* `TfNetworkTrainer` class
* `TfExperiment` class

Further features listed in #47 can be subsequently added, but are not essential core functionality.
Note that tests will need to be supressed until [tensorboardlogger is released for trixi](https://github.com/MIC-DKFZ/trixi/issues/12).


justusschock:
@ORippler I had to remove the tensorflow tests and installation completely in CI/CD. For using tensorflow backend, you have to install [trixi](https://github.com/MIC-DKFZ) from `develop` branch

PR #63, opened 2019-02-20
re-enable model kwargs

### LINKED ISSUE #59: [FeatureRequest] kwargs for models
The previous versions of delira had a kwargs argument for models. This was quite useful, it would be great if it would be available in the version of delira as well.

justusschock:
Re-Introduce model kwargs and fix #59 

@ORippler : what happens to the variable parameters if this is a range (e.g. for hyperparameter tuning)? Will we assume, that only sampled values will be given to class instantiation?


ORippler (approved):
looks good to me. As i do not know how right now how tune performs the trial runs it might need to be refactored in future

PR #66, opened 2019-02-25
Trainer refactoring

### LINKED ISSUE #39: [FeatureRequest] Add Predictor class
Add `Predictor`-API

> Scheduled after #17 to avoid refactoring

### LINKED ISSUE #46: [FeatureRequest] Merge TfExperiment and PytorchExperiment
**Description**
Merge TfExperiment and PytorchExperiment to remove code duplication. Seemingly only the handling of `tf.Graph` objects has to be shifted into the Trainer or alternatively the TfNetwork

**Proposal**
Add specific `tf.Graph` namespaces to TfNetwork by providing them to `tf.Session`. This should eliminate the need to reset the global default graph per Trainer instance.

justusschock:
This is a first draft to refactor trainer (combine code where possible) and introduce a predictor.

The metric logging was moved from the networks closure to the trainer.

@ORippler : maybe we could also merge experiments and rename the `AbstractTrainer` to `BaseTrainer` since it is not abstract anymore?

Also some tf tests fail due to shape missmatch. Have you any idea why?

@mibaumgartner : does this match your idea of a predictor? Do you have any improvements compared to your own prototype?

fix #39 
fix #46 ?

EDIT: Docstrings are still missing


delira/training/predictor.py
```diff
@@ -0,0 +1,208 @@
...
+
+        assert mode in ['highest', 'lowest'], "Invalid Comparison Mode"
+
+        if mode == 'highest':
+            return new_val_score > old_val_score
+        elif mode == 'lowest':
+            return new_val_score < old_val_score
+
+    @staticmethod
+    def calc_metrics(groundtruths, *predictions, metrics={}, metric_keys=None):
```
mibaumgartner: `calc_metrics` is called in the network trainer without the option to change the metrics_keys. 

https://github.com/justusschock/delira/blob/722cd55dd308f2ff067f34d5eb323511fe3073c2/delira/training/abstract_trainer.py#L334-L338

It may be useful to introduce a new argument in the trainer to set the metrics_keys so it is not always necessary to overwrite the function if different metric keys are needed.

delira/training/predictor.py
```diff
@@ -0,0 +1,208 @@
...
+                preds = self.predict(batch_dict)
+
+                predictions_all.append(preds)
+                inputs_all.append(batch_dict)
+
+                batch_list = []
+
+        batchgen._finish()
+
+        predictions_all = zip(*predictions_all)
```
mibaumgartner: For `zip` to work properly it is necessary that `predictions_all` is a list of lists. For some networks (like detection networks) which complete multiple tasks (detection, segmentation, classification) I like to use a dict for my network output (so I don't have to remember which entry belongs to which task). Is there any possibility to allow that here?

delira/training/abstract_trainer.py
```diff
@@ -1,31 +1,57 @@
...
+                       "output_device",
+                       "_callbacks"]
+
+    def __init__(self,
+                 network: AbstractNetwork,
+                 save_path: str,
+                 losses: dict,
+                 optimizer_cls: type,
+                 optimizer_params: dict,
+                 train_metrics: dict,
```
mibaumgartner: Maybe allow `train_dataset_metrics`?

justusschock: This was not allowed on purpose, because this would need the whole trainset to be cached, which is often not a good idea.

delira/training/predictor.py
justusschock: We would have to entirely switch to a dict, because maintaining both possibilities would be an unnecessary overhead I guess.

What do you think @haarburger @ORippler ?

delira/training/metrics.py
```diff
@@ -1,106 +1,37 @@
...
 
-        .. warning::
-            :class:`AurocMetricPyTorch` will be removed in next release
+        for k, v in arrays.items():
+            arrays[k] = v.reshape(v.shape[0], -1).astype(np.uint8)
 
-        """
-        def __init__(self):
-            super().__init__()
+        return self._score_fn(**arrays, **kwargs)
```
mibaumgartner: The AUC score would be a nice addtional feature. If needed, I can provide an implementation later today or tomorrow.

delira/training/abstract_trainer.py
mibaumgartner: For classification and detection it is only needed to cache the label and bounding boxes, which is pretty reasonable. Nevertheless, I get why it's not in there but I think it would be nice to have the option.

delira/training/metrics.py
justusschock: Feel free to do so.

ORippler: @mibaumgartner please resolve if done or no longer planned

delira/training/predictor.py
```diff
@@ -0,0 +1,208 @@
...
+
+        # check if key has been set once
+        if key in self.__KEYS_TO_GUARD and hasattr(self, key):
+            raise PermissionError("%s should not be overwritten after "
+                                  "it has been set once" % key)
+        else:
+            super().__setattr__(key, value)
+
+    @staticmethod
+    def _is_better_val_scores(old_val_score, new_val_score,
```
ORippler: from what this function does, it should live in the `AbstractTrainer` and not in the predictor imo. It is so far also only used by the `AbstractTrainer`

delira/training/predictor.py
```diff
@@ -0,0 +1,208 @@
...
+        if isinstance(data, dict):
+            data = [data[key] for key in keys]
+
+        return self._convert_batch_to_npy_fn(
+            self.module(
+                *data
+            )
+        )
+
+    def predict_data_mgr(self, datamgr, batchsize=None, verbose=False):
```
ORippler: in tf, a singleton dimension is added somewhere in here, leading to shape missmatches for inputs

ORippler (changes requested):
will investigate shape errors tomorrow. Furthermore, experiment.stratified_kfold_predict has to be adapted (might also be related to the shape errors)

delira/training/predictor.py
ORippler: Singleton is also added for pytorch presumably, as `classification_pytorch.ipynb` fails with the following error message:

```RuntimeError: Expected 4-dimensional input for 4-dimensional weight [64, 1, 7, 7], but got 5-dimensional input of size [64, 1, 1, 224, 224] instead```

We again have a singleton dimension (and some additional shape missmatch) when calling `predictor.predict_data_mgr`
Real question is though: why are our tests not catching this?

File: delira/training/predictor.py [hunk 7]
```diff
@@ -0,0 +1,208 @@
...
+                batch_dict = {}
+                for batch in batch_list:
+                    for key, val in batch.items():
+                        if key in batch_dict.keys():
+                            batch_dict[key].append(val)
+                        else:
+                            batch_dict[key] = [val]
+
+                for key, val_list in batch_dict.items():
+                    batch_dict[key] = np.array(val_list)
```
ORippler: why this got changed from [np.concatenate](https://github.com/justusschock/delira/blob/b663d982d87f3a0be7b346d25576770b184d4db9/delira/training/tf_trainer.py#L463) to np.array? should fix the singleton issue presumably:

delira/training/predictor.py
ORippler: I think we shouldn't mix `lists` and `dicts`. In current master, we only use lists for `labels_all` and `outputs_all`, but now we intermingle it already now with `predictions_all` and `total_input_dict`. If we need the whole information in `total_input_dict` in order to calculate proper validation metrics, we should switch to dicts totally.

delira/training/predictor.py
justusschock: My mistake. Will fix this probably tomorrow

delira/training/metrics.py
mibaumgartner: Sorry, I didn't get any notification. I already sketched a first version, gonna test it during the next days (maybe even today) and then create a pull request.

delira/training/abstract_trainer.py
justusschock: @ORippler do you think, we should have this option for the trainset as well?

delira/training/predictor.py
justusschock: I think, I'd like to move this to another issue, since this issue shouldn't change anything on the user-side, but switching from lists to dicts will.

justusschock:
@ORippler why did you force push and delete multiple functions?

ORippler:
@justusschock I worked on the wrong branch and merged + pushed the master by accident. Force push should have reverted to original state.

ORippler (changes requested):
Okay tests seem to be passing so far. Further ToDOs/Open Questions:

* [x] Write/Update Docstrings for all classes
* [x] Write Unittests for `Predictor` class
* [x] Discuss/decide on whether conversion from list of list to list of dicts should be part of this PR or another one
* [x] Argue about the usability of training dataset metrics

delira/training/abstract_trainer.py
ORippler: What benefit would `train_dataset_metrics` bring us?

justusschock:
To move this on, I'll try to write the tests and add the docs this week. 

IMO the list/dict conversion should be separated and both dataset-metrics (training & test) should not be integrated here, because they'd break the lazy behavior of the lazy datasets. 
At least we need to check, if we're using lazy datasets and only use dataset-metrics if we don't use them, but I'd prefer to remove them until they are fully supported (in a lazy and non-lazy way). 

Any thoughts? @ORippler @mibaumgartner

justusschock:
@ORippler tests for torch are passing on my machine, but strangely tests for tf aren't can you maybe have a look at that?

despite that I think I finished the trainer. I still have to update the class entries in the doc files once all tests are passing

ORippler:
so i still get 2 errors for the Predictor. In the experiment, i get shape mismatch, and for the [predictor.predict on a single element complaining about float, the default function returns a tuple instead of a dict with identical mapping](https://github.com/justusschock/delira/blob/5ca4024691b56a91a63ff493ea3e14359ddcc59b/delira/training/predictor.py#L28).

ORippler:
so after discussing with @justusschock we decided to include the change from lists to dicts in this PR. In this process, we will also change the Predictor class to work based on dicts, with predictor.predict() utilizing keys to specify inputs and returning a dict. Thought's on this @haarburger @mibaumgartner ?

justusschock:
@ORippler I think I adressed most of the issues to switch to dicts. Unfortunately I don't know, where to apply the switch inside the Tf-Networks `__call__` (I'm afraid to actually destroy parts of the graph if wrapping at the wrong place). Can you maybe have a look on that?

justusschock:
8c8c9cd002f67cb0e3bc96e46829c7992622d9fa should fix #46 and #86 

ToDos remaining:

- [x]  Update Tests
- [x] Update Docstrings
- [ ] Check Registration of all classes in doc-files

delira/training/experiment.py
```diff
@@ -1,388 +1,582 @@
...
+    * The K-Fold logic (including stratified and random splitting)
+
+    * Argument Handling
 
     """
-    @abstractmethod
-    def __init__(self, n_epochs, *args, **kwargs):
+
+    def __init__(self,
+                 params: Parameters,
```
mibaumgartner: Is it possible to make the `params` object optional in the `__init__`? (It would change the signature and thus backwards compatibility could be problematic)

delira/training/experiment.py
```diff
@@ -1,388 +1,582 @@
...
+            during actual training
+        name : str or None
+            the Experiment's name
+        save_path : str or None
+            the path to save the results and checkpoints to.
+            if None: Current working directory will be used
+        key_mapping : dict
+            mapping between data_dict and model inputs (necessary for
+            prediction with :class:`Predictor`-API)
+        val_score_key : str or None
```
mibaumgartner, line 73: I think  it's more intuitive if `val_score_key` (currently set in `__init__`) and `val_score_mode`(currently set in `run`) would be set in the same function (maybe in `run`). What do you think about this?

delira/training/tf_trainer.py
```diff
@@ -60,55 +85,52 @@ def __init__(self, network, save_path,
...
 
         self._setup(network, optim_fn, optimizer_cls, optimizer_params,
-                    lr_scheduler_cls, lr_scheduler_params, gpu_ids)
+                    lr_scheduler_cls, lr_scheduler_params,
+                    key_mapping, convert_batch_to_npy_fn, gpu_ids)
 
         for key, val in kwargs.items():
             setattr(self, key, val)
 
     def _setup(self, network, optim_fn, optimizer_cls, optimizer_params,
```
mibaumgartner, line 131: It seems like the `PytorchNetworkTrainer` already supports the `resume` function of the `experiment` class and loads the latest checkpoint. Currently, it is a bit misleading if only one trainer supports this functionality. In general, I'm a big fan to have the option to resume the training.

mibaumgartner:
https://github.com/justusschock/delira/blob/e1189b7fdd2808e280b70b05ee27fe422d32cb66/delira/training/pytorch_trainer.py#L323-L348
Looks like a code duplication with
https://github.com/justusschock/delira/blob/e1189b7fdd2808e280b70b05ee27fe422d32cb66/delira/training/base_trainer.py#L222-L247

delira/training/experiment.py
justusschock: I think I left it that way, because I think this is what it was like until now, but I agree

delira/training/experiment.py
justusschock: We would have to change the argument order here and I'm not sure if I'd like that, because IMO, the parameters are more important than other stuff and I can't really imagine a scenario where you wouldn't pass parameters here. What do you think @ORippler?

justusschock:
I missed this one, I'll remove it

delira/training/tf_trainer.py
ORippler: we should imo include a switch on wether to restore from best model or latest model.
Also, for `TfNetworkTrainer`, `start_epoch` still has to be updated accordingly

delira/training/experiment.py
ORippler: I also don't know of a situation where you wouldn't pass params to `__init__`. Though I think we should rethink/define on what to pass as additional kwarg and what to include inside the params.

Idea: Everything that can be tuned/varied during training/testing should go into params, as they are intended to be utilized for the hyperparameter search)

delira/training/tf_trainer.py
justusschock: I think the resume inside the trainer should only restore the latest model, since we want t continue training here...

delira/training/experiment.py
mibaumgartner: I encountered multiple situations where I wanted to initialise the experiment to get the `savepath` without initialising the parameters, but maybe that is only me ^^

ORippler: with current implementation, it must be typing.Union[str, Parameters]

delira/training/experiment.py
```diff
@@ -1,388 +1,583 @@
...
+                                         params.nested_get("num_epochs"))
+
         super().__init__(n_epochs)
-        self._run = 0
 
-    @abstractmethod
-    def setup(self, *args, **kwargs):
-        """
-        Abstract Method to setup a :class:`AbstractNetworkTrainer`
+        # params could also be a file containing a pickled instance of parameters
```
ORippler: if we accept filepaths for params, this must come before handling `n_epochs is None`

delira/training/predictor.py
```diff
@@ -0,0 +1,277 @@
...
+    --------
+    :class:`PyTorchNetworkTrainer`
+
+    """
+
+    # static variable to prevent certain attributers from overwriting
+    __KEYS_TO_GUARD = []
+
+    def __init__(self, model, key_mapping: dict, 
+                 convert_batch_to_npy_fn=lambda x: x,
```
ORippler: lambda x: x no longer captures the desired default behaviour (which seems to be [discard *args and pass **kwargs) of returned model.call()](https://github.com/justusschock/delira/blob/trainer_refactoring/delira/training/predictor.py#L113). Either make it positional arg or change default setting. Also, the naming is no longer fitting (we convert the tensors inside the dict entries to np.ndarrays if required).

delira/training/tf_trainer.py
ORippler: okay, still we should update `start_epoch` then

delira/models/classification/classification_network_tf.py
```diff
@@ -181,11 +188,7 @@ def closure(model: typing.Type[AbstractTfNetwork], data_dict: dict,
...
 
-        for key, val in {**metric_vals, **loss_vals}.items():
-            logging.info({"value": {"value": val.item(), "name": key,
-                                    "env_appendix": "_%02d" % fold
-                                    }})
 
         #logging.info({'image_grid': {"image_array": inputs, "name": image_names,
         #                             "title": "input_images", "env_appendix": "_%02d" % fold}})
-        return metric_vals, loss_vals, [preds]
+        return metric_vals, loss_vals, {"pred": preds}
```
ORippler: is this still required? As far as i can see, the dict structure is only required to pass it through Predictor._convert_batch_to_npy(). My outputs are now naturally of dict type and therefore I would pass the whole returned dict through here.

delira/training/predictor.py
```diff
@@ -0,0 +1,277 @@
...
+
+        Returns
+        -------
+        dict
+            dict with metric results
+        """
+        if metric_keys is None:
+            metric_keys = {k: ("pred", "label") for k in metrics.keys()}
+
+        return {key: metric_fn(batch_dict[metric_keys[key][0]],
```
ORippler: we should switch to allow variable length for metric_keys

delira/training/predictor.py
```diff
@@ -0,0 +1,277 @@
...
+
+        # converts positional arguments and keyword arguments,
+        # but returns only keyword arguments, since positional
+        # arguments are not given.
+        return self._convert_batch_to_npy_fn(
+            **pred
+        )[1]
+
+    def predict_data_mgr(self, datamgr, batchsize=None, metrics={},
+                         metric_keys=None, verbose=False):
```
ORippler: Should we add the ability to compute metrics over the whole data_mgr? Instead of returning a list of batch_level metrics in addiction to all predictions. I think we had decided not to for training data_managers...
Also, description of what is returned is missing

ORippler (changes requested):

delira/training/experiment.py
justusschock: will fix this

delira/models/classification/classification_network_tf.py
justusschock: No, this isn't required, This was only my first try to make it work with TF :D

delira/training/predictor.py
justusschock: We had decided to do this in general but to postpone this, until we have a way to do this lazy and non-lazy, because for large datasets (e.g. imagenet) we might also break the lazy behavior with the validation/testset.

Will update the docstrings tomorrow together with your other comments

delira/training/predictor.py
```diff
@@ -0,0 +1,286 @@
...
+                    for key, val in batch.items():
+                        if key in batch_dict.keys():
+                            batch_dict[key].append(val)
+                        else:
+                            batch_dict[key] = [val]
+
+                for key, val_list in batch_dict.items():
+                    batch_dict[key] = np.concatenate(val_list)
+
+                preds = self.predict(batch_dict)
```
ORippler: do we allow nested dicts? or not

justusschock: you mean as input? This shouldn't matter, as long as your network/`prepare_batch` handle this correctly. 

For the output this becomes problematic, since we do some concatenation based on these values and we can't concatenate dicts. Another possibility would be to flatten these dicts by something like this:

```python
import collections

def flatten(d, parent_key='', sep='_'):
    items = []
    for k, v in d.items():
        new_key = parent_key + sep + k if parent_key else k
        if isinstance(v, collections.MutableMapping):
            items.extend(flatten(v, new_key, sep=sep).items())
        else:
            items.append((new_key, v))
    return dict(items)
```

This would be a design decision, we'd have to make now (as this currently causes the tests to fail). I don't really have an opinion on this. What do you think @ORippler @mibaumgartner @haarburger ?

ORippler: Yeah i also identified that as the reason for failure.

I dislike the idea of flattening the dict, as the new keys would need to be passed to `calc_metrics` and the user informed of this.

I would prefer a recursion-based solution to allow for nested dicts, though `calc_metrics` would also need to be updated (currently, the values used for metric calculation need to lie on the top-level of the dict).

mibaumgartner: I'm not sure if it is beneficial to support nested dicts. Are there any useful scenarios where this could/must be used?

justusschock: I wouldn't like a recursion based approach. I can't see any benefits from it, but if we need to resolve arbitrary recursions multiple times this is bad style IMHO

ORippler: @mibaumgartner the reason we run into that failure is because [tf.Session.run()](https://www.tensorflow.org/api_docs/python/tf/Session#run) allows nested dicts and i used that functionality already inside `ClassificationNetworkBase` (it makes the logging of training loss easier as we have model.closure as static function for pytorch's multi-gpu approach). Could handle the unpacking inside the classes' `__call__` though.

justusschock: @ORippler I'm pro nested dict, if we can do this by resolving the recursion only once (i.e. something similar like flattening). What I don't like: If we would have to resolve the recursion mutiple times (i.e. one time for concatenation and another time for metric calculation (and maybe a third time for logging/reporting back to HP search in the future). This gets extremly inefficient quite fast.

Another problem would be, that we could have the same key in multiple subdicts and it is unclear which item the key refers to (e.g. in metric calculation).

justusschock: @mibaumgartner @ORippler I would like to move this on. What are we gonna do here?

mibaumgartner: If it is benficial for tf applications, we should implement a soultion....

justusschock: But which one? The only one, I can think of, resolving the possible ambiguity in keys would be the one above. Alternatively we could wrap in in a `LookupConfig` (the parameters baseclass) and use `nested_get` here as well

justusschock: So, I just implemented the version with `LookupConfig` and nested dicts.

justusschock:
There are only two issues left:

-  The `test_experiment_test_tf` functions seems to be unable to find the `dense_2` layer (which does not make sense to me)
-  `trixi` still has a problem with their dependencies, which is, why we had to downgrade the scipy version to 1.0.0, until they have released a package containing a fix for [this](https://github.com/MIC-DKFZ/trixi/issues/15). The problem is, that scipy 1.0.0 is not available for python 3.7. So for now the python 3.7 tests will be failing until the new trixi release.

I think, once the first issue is fixed, I will look to update all examples and documentation and then this is ready to merge; Any other opinions? @haarburger @ORippler @mibaumgartner ?

justusschock:
I just added two things to the `Predictor`:

* The conversion of the batchdict from native tensors to numpy (since the batchdict may be converted inplace during `predict`)
* A `as_generator` option, which yields the results of each batch instead of adding them to the list and returning the whole list afterwards. This should be useful if we only want to predict, since for huge datasets (especially if 3D) tasks like segmentation might cause `OutOfMemoryError`s even if we only store predictions in memory and this way they can be processed/saved on a per-batch basis

EDIT:
I failed with the generator part, but fixed it now... The problem with generators is, that every function, that contains a yield in it's body returns a generator name (even if the actual yield is never triggered). This prevents us from using a solution which either returns a generator or the complete predictions and metrics. Because I think, we should definitely support this kind of behavior, https://github.com/justusschock/delira/pull/66/commits/b510ffb54cc8b93e4c1aa1623c5196d7a2df5071 makes the `predictor.predict_data_mgr ` always return a generator object, which is of size 1 if the `lazy_gen` flag is False. 
https://github.com/justusschock/delira/pull/66/commits/13d3adbfd5da69db0fc821e62456fa4cfbfa8e3d updates the `experiment.test` function accordingly

justusschock:
I think we're almost good to go. The only remaining issue is the one with tensorflow not finding some resources in our tests. According to [this](https://github.com/tensorflow/tensorflow/issues/28287) issue it is most likely due to a thread that is spawned anywhere... Any Ideas on why this happens now and didn't happen before?

justusschock:
The trainer is now good to be merged. The only failing test is python 3.7 which is due to trixi dependencies but completely unrelated to this PR

mibaumgartner (approved):
Implementation looks good :+1: 
We should include the `Predictor` inside the `training.rst` so it is displayed correctly in the documentation.
I think that should be a quick fix, so i approve the changes.

justusschock:
The file already existed, I just forgot to include it into the root file. Done now

Issue #67, opened 2019-02-28
[FeatureRequest] Dataset use w/o ground truth  


gedoensmax:
**Description**
It would be nice to be able to use any Dataset without any given ground truth.

**Feature History**
I tried not setting gt_extensions and giving it an empty list.
Both did not work and i would prefer if an empty list would be the default value.

**Proposal**
There would have to be a check in the `_is_valid_image_file` method if there are file extensions given.
If not the image should be valid if it is existent.

**Additional context**
This could be a solution:
```
…
has_label = not self._gt_extensions
for ext in self._gt_extensions:
label_file = fname.rsplit(".", maxsplit=1)[0] + ext
if os.path.isfile(label_file):
     has_label = True

return is_valid_file and has_label
```

justusschock:
This seems to be a good idea, since you already figured out a solution, would you send a PR?

justusschock:
Fixed by #68

PR #68, opened 2019-02-28
[FeatureRequest] Dataset use w/o ground truth

gedoensmax:
Minor changes to dataset.py so that it can now create unlabeled datasets.


justusschock:
should fix #67

justusschock (changes requested):
There are still a few things not working here.

It'd be great if you could have a look at the currently failing tests, otherwise I'll do that tomorrow.

justusschock (approved):

PR #72, opened 2019-03-12
Increment tf version

justusschock:
Increments tf version to 1.13.1 and finally supports python3.7, which means, we can enable our tf-tests.

@ORippler I don't know, where these test failures are coming from, the code seems alright for me and works seems to work on my local machine.


ORippler (changes requested):
@justusschock I managed to fix the tf.Graph pollution issue that caused the io tests to fail. We still have failing test for experiment_torch that needs to be adressed and some tests take so long that travis sometimes aborts builing. Also, model tests are skipped due to the memory constraint i presume

justusschock:
@ORippler the failure in torch experiment is strange since this PR does not change anything on this, and it did not fail before. 

Regarding the duration of the tests: I would move the functional tests of the experiment to separate functions, since this probably cuases the abortion.

The model tests are indeed skipped due to the memory constraint (we only have 7.5 GB here)

ORippler:
@justusschock tf tests are now all passing

justusschock:
@ORippler Torch tests should also be passing right now.

ORippler (approved):

ORippler:
incrementation was done by #79 
branch will be used to add further unittests

PR #76, opened 2019-03-14
Random sampler

mibaumgartner:


delira/data_loading/sampler/weighted_sampler.py
```diff
@@ -1,7 +1,8 @@
-from ..dataset import AbstractDataset
-from .abstract_sampler import AbstractSampler
```
justusschock: replace absolute imports with relative imports

justusschock (changes requested):

justusschock (approved):

PR #79, opened 2019-03-21
Switch test framework

### LINKED ISSUE #78: [Bug][URGENT] Fix Unittests
Pytest randomly freezes in CI/CD -> switch to other test framework

justusschock:
Attempt to fix #78 

Should also be more natively integrated to python because `unittest` is the python-integrated framework for testing.


ORippler (approved):
lets see if it works more stable now

Issue #83, opened 2019-04-04
[FeatureRequest] DatasetIterator


justusschock:
Make the dataset iterable (probably sufficient to simply add a `__curr_index` attribute and  `__next__`  and `__iter__` functions to `AbstractDataset`)

justusschock:
@mibaumgartner can you maybe include this to your dataset refactoring?

justusschock:
I just rethought this, and I'd prefer to have an explicit `_DatasetIter` class, which has to be initialized and returned in dataset's `__iter__`, since I don't like the idea to mix up the concept of a dataset (which is only for loading and processing data) and an iterator, which holds an internal state.

The `_DatasetIter` class could probably look something like this:
```
class _DatasetIter(object):
    def __init__(self, dset):
        self._dset = dset
        self._curr_index = 0

    def __iter__(self):
       return self

    def __next__(self):
        if self._curr_index > len(self._dset):
            raise StopIteration
        
        sample = self._dset[self._curr_index]
        self._curr_index += 1
        return sample
```

while the dataset's `__iter__` would look something like this:
```
def __iter__(self):
    return _DatasetIter(self)
```

I think, this should be the most elegant and still quite simple way to do this.

justusschock:
done by #88

PR #88, opened 2019-04-16
Dataset refactoring

mibaumgartner:
Are there any additional functions which should be added or changes?


delira/data_loading/load_utils.py
```diff
@@ -1,8 +1,47 @@
 import numpy as np
+import os
+import collections
 from skimage.io import imread
 from skimage.transform import resize
 
 
+def norm_range(data):
```
justusschock: Maybe we want to add a range argument here, for being able, to use the same function for normalization between [0, 1]?

delira/data_loading/load_utils.py
```diff
@@ -41,3 +80,137 @@ def default_load_fn_2d(img_file, *label_files, img_shape, n_channels=1):
...
+
+
+class LoadSample:
+    def __init__(self,
+                 sample_ext: dict,
+                 sample_fn: collections.abc.Callable,
+                 dtype={}, normalize=[], norm_fn=norm_range,
+                 **kwargs):
+        """
+        Provides a callable to load a single sample fom multiple files in a
```
justusschock: typo

delira/data_loading/dataset.py
```diff
@@ -233,167 +225,134 @@ def __len__(self):
...
+        else:
+            # call _sample_fn for all elements inside directory
+            assert os.path.isdir(path), '%s is not a valid directory' % dir
+            for p in tqdm(os.listdir(path), unit='samples',
+                          desc="Loading samples"):
+                data.append(self._load_fn(os.path.join(path, p),
+                                          **self._load_kwargs))
         return data
 
-    def _is_valid_image_file(self, fname):
```
justusschock, line 307: Maybe we want to include this (or something similar) as a helper function in the utilities until the next major release?

delira/data_loading/dataset.py
```diff
@@ -407,76 +366,121 @@ def _make_dataset(self, path):
...
-            is valid data sample
+        dict
+            loaded data sample
+        """
+        data_dict = self._load_fn(self.get_sample_from_index(index),
+                                  **self._load_kwargs)
+        return data_dict
+
+
+class BaseExtendCacheDataset(BaseCacheDataset):
```
justusschock, line 436: I quite don't get the difference of `BaseCacheDataset` and `BaseExtendCacheDataset`. From the docstring they should be the same?

delira/data_loading/dataset.py
```diff
@@ -531,137 +535,10 @@ def __len__(self):
         return sum([len(dset) for dset in self.data])
 
 
-class Nii3DLazyDataset(BaseLazyDataset):
```
justusschock: I guess we should keep this for the next minor release (along with a deprecation warning and adapted if necessary) and remove it in the next major release

delira/data_loading/dataset.py
```diff
@@ -771,10 +648,10 @@ def __getitem__(self, index):
 
             data = self.data[index]
             data_dict = {"data": np.array(data[0]),
-                         "label": data[1].numpy().reshape(1).astype(np.float32)}
+                         "label": data[1].reshape(1).astype(np.float32)}
```
justusschock, line 804: Just to remember: This will fail with torchvision < 0.2.2; fixed in #66

justusschock (changes requested):
Looks good in general, I left a few comments, but nothing major, except the `_DatasetIterator` from #83 is still missing

delira/data_loading/load_utils.py
mibaumgartner: How would you implement this feature? The `load_sample` functions gets a function, so the user has the ability to use any normalisation function.
Maybe something like that?

```python
def norm_range(mode):
    def norm_fn(data):
        # do something depending on mode
    return norm_fn
```

delira/data_loading/dataset.py
mibaumgartner: The `BaseCacheDataset` uses the `append` to add elements to the internal data list. `BaseExtendCacheDataset` uses `extend`, so the load_fn is expected to return an iterator which is added to the internal data list -> this should enables loading of multiple samples at once
I'm going to change the doc string so it is more clear to everybody.

delira/data_loading/dataset.py
mibaumgartner: I added a version of the Lazy and Cache version. `_img_extensions` and `_label_extensions` seemed to have no functionality, so they are not used anymore.

delira/data_loading/dataset.py
mibaumgartner: Do you have a concrete idea where/how I could add this? The new datasets do not use `_img_extensions` as they were previously implemented.

justusschock: I'd just leave them as a variable so they have to be passed explicitly.

delira/data_loading/load_utils.py
justusschock: I think, this is the most suitable way to implement this. I thought about additionally passing an explicit range, but this way we would not have a uniform API

delira/data_loading/load_utils.py
```diff
@@ -63,6 +64,34 @@ def norm_zero_mean_unit_std(data):
     return (data - np.mean(data)) / np.std(data)
 
 
+def _is_valid_image_file(self, fname):
```
justusschock: I think, we should deprecate this too

delira/data_loading/load_utils.py
```diff
@@ -63,6 +64,34 @@ def norm_zero_mean_unit_std(data):
...
+    ----------
+    fname : str
+        filename of image path
+     Returns
+    -------
+    bool
+        is valid data sample
+     """
+    is_valid_file = False
+    for ext in self._img_extensions:
```
justusschock: this won't work, you have to pass the extensions explicitly to this function

delira/data_loading/load_utils.py
```diff
@@ -63,6 +64,34 @@ def norm_zero_mean_unit_std(data):
...
+        if fname.endswith(ext):
+            is_valid_file = True
+
+    has_label = not self._gt_extensions
```
justusschock: same as for `img_extensions`

justusschock (approved):

PR #92, opened 2019-05-02
Fix tf resnet18

### LINKED ISSUE #91: [Bug] Tf ResNet18 is missing Residual Connections
**Description**

Forgot to include Residual Connecitons into Tf ResNet18 implementation

ORippler:
Should fix #91 . Implementation should now follow [Kamga et al. 2015](https://arxiv.org/abs/1512.03385), though I don't know why they did a 1x1 convolution with stride 2.


justusschock (approved):
Looks good to me. Should we also wrap this in the backend guard or won't we do this, because it is never directly used anywhere?

delira/models/classification/ResNet18.py
```diff
@@ -17,53 +18,91 @@ def __init__(self, num_classes=None):
...
         self.conv5_1 = conv2d(filters=512, strides=2, kernel_size=3,
                               padding='same')
+        self.batchnorm5_1 = batchnorm2d(axis=1)
         self.conv5_2 = conv2d(filters=512, strides=1, kernel_size=3,
                               padding='same')
+        self.batchnorm5_2 = batchnorm2d(axis=1)
 
-        self.gap = gap2d()
         self.dense1 = dense(1000)
         self.dense2 = dense(num_classes)
```
mibaumgartner: This implementation would lead to a different network structure thanin torchvision. I think there should be a single fully connected layer with the correct number of classes (this would be consistent with torchvision).

mibaumgartner:
The number of residual blocks is wrong. Per "resolution" there should exist 2 residual blocks (in ResNet18), here only a single block exists (resulting in an ResNet10).

delira/models/classification/ResNet18.py
justusschock: Although I think that the one in torchvision is not the original one proposed in the paper

mibaumgartner: The original paper only refers to image net classification (if I'm not mistaken) and thus only covers the case with 1000 classes. In my opinion, it would make sense to keep the structure with one fully connected layer (as stated in the paper) and only reduce the number of output classes.
What are the differences between torchvision and the original paper?

ORippler: I think we should stick to the paper-version (which is identical to the torchvision implementation and consists of 1 final fc layer)

ORippler:
> 
> 
> The number of residual blocks is wrong. Per "resolution" there should exist 2 residual blocks (in ResNet18), here only a single block exists (resulting in an ResNet10).

Yes you are right, I also forgot to utilize the `training` bool for the normalization layers. will revisit this tomorrow

mibaumgartner (approved):

ORippler:
> Looks good to me. Should we also wrap this in the backend guard or won't we do this, because it is never directly used anywhere?

I think it's cleanest if we apply the guard only where necessary (e.g. inside the `__init__.py` files and if we have different backends inside the same `.py` file), but I can also add the guards if you prefer it that way

delira/models/classification/ResNet18.py
```diff
@@ -6,68 +6,133 @@
...
+
+
+def get_image_format_and_axis():
+    image_format = tf.keras.backend.image_data_format()
+    if image_format == "channels_first":
+        return image_format, 1
+    elif image_format == "channels_last":
+        return image_format, -1
+    else:
+        raise Exception("Image format type unknown")
```
justusschock: Please make this either `RuntimeError` or `ValueError`... Using bare Exceptions is not recommended, since we cannot explicitly catch this type of exception (if we'd like to for some reason) and ignore all the others.

EDIT: Maybe you could also include the returned format?

justusschock (approved):
I approve this, because there is only 1 minor issue, which does not prevent the code from working

mibaumgartner (approved):

delira/models/classification/ResNet18.py
ORippler: adressed in 5b55552af1b8b30dcb7c7af257bb34270cc6c5bb

justusschock (approved):

justusschock (approved):

PR #94, opened 2019-05-11
Make Parameters copyable

mibaumgartner:
Make Parameters copyable and check permutation before updating


justusschock (approved):
Can be merged once #95 was merged to ensure CI/CD running again

PR #97, opened 2019-05-14
Move to new APEX.amp API

### LINKED ISSUE #71: Move to new AMP API
https://nvidia.github.io/apex/amp.html#transition-guide-for-old-api-users

justusschock:
Closes #71 

Re-Enables current (unified and more stable) AMP API.

Apex must still be installed manually and following [this PR](https://github.com/NVIDIA/apex/pull/252) it's mixed precision training currently only works on GPU.

EDIT: Switched base branch to not intermingle this with trainer refactoring

### DIFF
--- a/delira/models/model_utils.py
+++ b/delira/models/model_utils.py
@@ -1,0 +1,26 @@
+import contextlib
+
+try:
+    # use apex loss scaling if possible
+    # (and enabled, this is done internally by apex)
+    from apex import amp
+except ImportError:
+    # use no loss scaling with same API if apex is unavailable
+    amp = None
+
+
+@contextlib.contextmanager
+def scale_loss(loss,
+               optimizers,
+               loss_id=0,
+               model=None,
+               delay_unscale=False):
+    if amp is None:
+        yield loss
+
+    else:
+        with amp.scale_loss(loss=loss, optimizers=optimizers,
+                            loss_id=loss_id, model=model,
+                            delay_unscale=delay_unscale) as _loss:
+            yield _loss
+    return
--- a/delira/models/model_utils.py
+++ b/delira/models/model_utils.py
@@ -14,13 +14,15 @@ def scale_loss(loss,
                optimizers,
                loss_id=0,
                model=None,
-               delay_unscale=False):
+               delay_unscale=False,
+               **kwargs):
     if amp is None:
         yield loss
 
     else:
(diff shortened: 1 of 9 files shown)


justusschock:
Failures are only due to TF tests and are irrelevant for this PR

delira/training/pytorch_trainer.py
```diff
@@ -52,9 +52,14 @@ def __init__(self,
                      metric_keys=None,
                      convert_batch_to_npy_fn=convert_torch_tensor_to_npy,
                      mixed_precision=False,
-                     mixed_precision_kwargs={"enable_caching": True,
-                                             "verbose": False,
-                                             "allow_banned": False},
+                     mixed_precision_kwargs={"opt_level": "o0",
```
mibaumgartner: Not sure if our default value should be `o0` because we have an additional `mixed_precision` argument. If I set `mixed_precision` to true, I would assume that some kind of mixed precision training is used and not full fp32 training.
I think we should either set the default value to `o1` (should be equivalent to the old implementation?) or rename `mixed_precision` to something more appropriate like `use_amp`.

I think, I would prefer the different default value.

delira/models/model_utils.py
```diff
@@ -0,0 +1,26 @@
...
+    # use apex loss scaling if possible
+    # (and enabled, this is done internally by apex)
+    from apex import amp
+except ImportError:
+    # use no loss scaling with same API if apex is unavailable
+    amp = None
+
+
+@contextlib.contextmanager
+def scale_loss(loss,
```
mibaumgartner, line 13: Maybe add `**kwargs` if the apex team decides to add some additional arguments in the future.

mibaumgartner (approved):
Left two comments which are a personal preference. The rest looks good to me 👍

delira/training/pytorch_trainer.py
justusschock: Changed by https://github.com/justusschock/delira/pull/97/commits/55d9ca898108e9acd99e1b651de336410dfeeec9

delira/models/model_utils.py
justusschock: Added by https://github.com/justusschock/delira/pull/97/commits/ea687b3411b5cd61ea49c455da263c40fd1ca3d8

PR #98, opened 2019-05-14
TorchScript backend

justusschock:
Adds `TorchScript` backend to delira.

ToDo:

- [x] Add Abstract Network
- [x] Add Trainer
- [x] Add I/O
- [x] Add Experiment
- [x] Add Tests
- [x] Add Tutorial
- [x] Add backend to Setup.py
- [x] Add backend to README.md
- [x] Add backend to getting started section of docs
- [ ] Add `forward` to API (must be postponed until [this PR](https://github.com/pytorch/pytorch/pull/20503) is officially released to support subclassing with `script_method` decorators in `torchscript`)

@mibaumgartner this is ready to review now. Do you have any plans for the eager execution backend for tf? Maybe we could make a backend branch, merge both backends to this branch and keep it in parallel until #66 is merged?

@haarburger I just requested a review from you, because you're also working with `PyTorch` and since you can export the whole graph in `TorchScript`, this might be interesting for you too.


mibaumgartner:
@justusschock I am planning an eager execution backend, but currently I do not have time for it (maybe in june). If anyone else has the capacity to do this earlier, I would be happy to review.

delira/io/torch.py
```diff
@@ -84,3 +82,85 @@ def load_checkpoint(file, **kwargs):
...
+        optimizers : dict
+            dictionary containing all optimizers
+        epoch : int
+            current epoch (will also be pickled)
+
+        """
+
+        # remove file extension if given
+        if any([file.endswith(ext) for ext in [".pth", ".pt", ".ptj"]]):
+            file = file.rsplit(".", 1)[0]
```
mibaumgartner: Maybe include a `logger.info` that the file extension is changed to `.ptj`.

delira/io/torch.py
```diff
@@ -84,3 +82,85 @@ def load_checkpoint(file, **kwargs):
...
+        OrderedDict
+            checkpoint state_dict
+
+        """
+        # remove file extensions
+        if any([file.endswith(ext) for ext in [".pth", ".pt", ".ptj"]]):
+            file = file.rsplit(".", 1)[0]
+
+        # load model
+        if os.path.isfile(file + ".ptj"):
```
mibaumgartner: I think we should not remove the file extension and only check for the `.ptj` ending. If no file with the `.ptj`ending is found, we should try to load the original file (extension was removed in line 143) and print some message.

delira/models/abstract_network.py
```diff
@@ -195,7 +197,84 @@ def __call__(self, *args, **kwargs):
...
+            Returns
+            -------
+            Any
+                result: module results of arbitrary type and number
+
+            """
+            return torch.jit.ScriptModule.__call__(self, *args, **kwargs)
 
         @staticmethod
         def prepare_batch(batch: dict, input_device, output_device):
```
mibaumgartner, line 281: Is it possible to script the `prepare_batch` so it is not required to rewrite it in C++ (if the model is deployed in c++ ^^) ?

delira/training/pytorch_trainer.py
```diff
@@ -504,3 +511,228 @@ def _update_state(self, new_state):
...
+            file_name : str
+                filename to save the state to
+            epoch : int
+                current epoch (will be saved for mapping back)
+            **kwargs :
+                keyword arguments
+
+            """
+            if file_name.endswith(".pt") or file_name.endswith(".pth"):
+                file_name = file_name.rsplit(".", 1)[0]
```
mibaumgartner, line 712: same as above

tests/training/test_experiments.py
```diff
@@ -261,6 +312,59 @@ def test_experiment_kfold_torch(self):
...
+                dset_test = DummyDataset(dataset_length_test)
+
+                dmgr_train = BaseDataManager(dset_train, 16, 4, None)
+                dmgr_test = BaseDataManager(dset_test, 16, 1, None)
+
+                exp.run(dmgr_train, dmgr_test)
+
+    @unittest.skipIf("TORCH" not in get_backends(),
+                     reason="No TORCH Backend installed")
+    def test_experiment_test_torchscript(self):
```
mibaumgartner, line 350: Maybe, we should add a test case where the experiment is resumed (this would implicitly test the `load_checkpoint` function, but maybe we should add an explicit one for this, too?)

justusschock: This shouldn't be necessary I think, because the last Checkpoint should be loaded after the end of `run`

delira/models/abstract_network.py
justusschock: I don't know if this is possible, because this way the `prepare_batch` could not be a static method any more I guess. Also I don't know if this would be beneficial since one would usually only use a scripted model to deploy it, where the prepare batch wouldn't probably be used directly.

PR #101, opened 2019-05-16
Add TF Eager Execution Backend

justusschock:
This Prototype aims to add support for TF Eager Execution.

I derived the `AbstractTfEagerNetwork` from `tf.keras.layers.Layer`, which is the base class for `tf.keras.Network`, which itself bases `tf.keras.Model`. It is the base class with minimal overhead, but still providing the functionality which has been used in the [official TF Eager Guide](https://www.tensorflow.org/guide/eager#top_of_page) and the [official TF Eager MNIST Example](https://github.com/tensorflow/models/blob/master/official/mnist/mnist_eager.py). I'm not sure, whether this is the best base class to use here, but it's the only suitable I found so far.

Tests are not passing (yet). Have to resolve some minor issues, but I'd be happy to get a first review from you (or even help resolving these issues).

ToDo:

- [x] resolve minor issues to make tests pass

- [ ] Add backend to Setup.py

- [ ] Add backend to README.md

- [ ] Add backend to getting started section of docs

- [ ] Add simple MNIST Example

EDIT: Tests for TF Eager backend are passing now, only tf graph backend tests are still failing (same as in #66 ). A working example of a simple network (only dense layers) can be seen in [these](https://github.com/justusschock/delira/blob/tf_eager_backend/tests/training/test_experiments.py#L162) tests. I'd pursue the remaining todos after #98 (and maybe #97 ) are merged, because #98 changes the requirements resolvement in `setup.py` which makes it easier to register new backends.


delira/io/tf.py
```diff
@@ -32,3 +36,51 @@ def load_checkpoint(file: str, model=None):
...
+
+def save_checkpoint_eager(file, model: AbstractTfEagerNetwork = None,
+                          optimizer: typing.Dict[str, tf.train.Optimizer] = None,
+                          epoch=None):
+    variable_list = _create_varlist(model, optimizer)
+
+    # can only save if variables exist, this is not the case if there was no
+    # input forwarded through the network (yet)
+    if variable_list:
+        saver = tf.contrib.eager.Saver(variable_list)
```
mibaumgartner, line 66: Is there any reason why we use `tf.contrib.eager.Saver` and not `tf.train.Checkpoint` as mentioned in the documentation ?

delira/training/tf_trainer.py
```diff
@@ -308,3 +314,304 @@ def load_state(self, file_name, *args, **kwargs):
...
+                network = tf.keras.utils.multi_gpu_model(
+                    network,
+                    gpu_ids,
+                    cpu_merge=True,
+                    cpu_relocation=False)
+
+                self.input_device = "/cpu:0"
+                self.output_device = "/cpu:0"
+            else:
+                self.input_device = "/gpu:0"
```
mibaumgartner: Does this always work? What if we want to train on `/gpu:1`?

mibaumgartner (changes requested):
Did a quick walkthrough and left some comments.

delira/io/tf.py
justusschock: There is one issue with it (if I remember correctly it was with saving multiple optimizer states, but I'm not completely sure what it was).

delira/training/tf_trainer.py
justusschock: You're right, I should query the first device in the list here.

PR #102, opened 2019-05-20
Additional backends

justusschock:
This PR will add additional backends (TorchScript in #98 and TF Eager Execution in #101 ) as well as the move to the new `apex.amp` API (#97 ).

Currently this PR is just open to track conversations and the Progress of the different Parts of it. It will be merged, after #66 , #97 , #98 and #101 are merged.

Progress so far:

- [x] Trainer Refactoring (#66) merged into `master`
- [X] new `apex.amp` API enabled (#97 )
- [X] `TorchScript` Backend (including new dependency resolution) (#98 ) merged 
- [X] `TF Eager Execution` Backend (#101 ) merged
- [X] `SkLearn` Backend (#109 ) merged
- [X] `chainer` Backend (#110 ) merged
- [x] Move backends to separate files -> 1 File / backend in `delira.training` including train-utils, trainer and experiment; 1 File / backend in `delira.models` including base network and other necessary stuff (like data parallel for `chainer`)
- [x] Refactor tests
- [x] Add defaults for `closure` and `prepare_batch` to each abstract network
- [x] Register all backends in __init__
- [x] Register all backends in setup

The following changes will be done in subsequent PRs.
* `MXNet` Backend (#119 ) merged
* Provide examples for all backends
* Add All backend-specific files to docs
* Add all backends to README
* Add all backends to Docs/Getting Started

Other possible backends include
* [Theano](http://deeplearning.net/software/theano/): Could be easy - quite similar to TF
* [Caffe2](https://caffe2.ai/): Probably won't integrate due to incorporation into PyTorch
* [CNTK](https://docs.microsoft.com/en-us/cognitive-toolkit/): Probably won't integrate because no longer actively developed and not that commonly used
* [PaddlePaddle (by Baidu)](https://github.com/PaddlePaddle/Paddle)

*If single backends of this list will be included, this should be done in separate issues/prs afterwards, since it is not yet clear which backends we want to integrate from here*


justusschock:
@mibaumgartner If you got some time, can you maybe have a look on why the tf eager stuff isn't working in CI/CD? It works, if I only run the tf eager tests on my local machine. I assume it's because we also have some tf graph stuff in the same test.

justusschock:
@mibaumgartner Can you maybe have a look at the failing Tf eager tests, once you got some time? 

A working minimal example is the following:

```
model = tf.keras.models.Sequential(
                layers=[
                    tf.keras.layers.Dense(64, input_shape=(
                        32,), bias_initializer='glorot_uniform'),
                    tf.keras.layers.ReLU(),
                    tf.keras.layers.Dense(
                        1,
                        bias_initializer='glorot_uniform')]
            )
import numpy as np
inputs = np.random.rand(1, 32)
labels = np.random.rand(1, 1)
with tf.GradientTape() as tape:
    preds = model(inputs.astype(np.float32))
    loss = tf.losses.absolute_difference(preds, labels.astype(np.float32))
    
grads = tape.gradient(loss, model.trainable_variables)
optim = tf.train.AdamOptimizer(learning_rate=1e-3)
optim.apply_gradients(zip(grads, model.trainable_variables))
```

Which basically is the same as done inside the [closure](https://github.com/justusschock/delira/blob/additional_backends/delira/models/backends/tf_eager/abstract_network.py#L111) too.

justusschock:
Current Problems are with `trixi` which introduced a PyTorch requirement even in `trixi-slim`, so we first need to integrate the new logging

justusschock:
@ORippler Tests are passing now. It would be great if you could make some time for a review

delira/models/backends/tf_eager/data_parallel.py
```diff
@@ -0,0 +1,48 @@
...
+            the module to scatter across different devices
+        devices : list
+            list of ints specifying the GPU indices
+        """
+        super().__init__()
+
+        self._closure = module.closure
+        self._prepare_batch = module.pepare_batch
+
+        self.module = tf.keras.utils.multi_gpu_model(module, devices, True)
```
ORippler, line 26: This is most likely not going to work (it only works when using high-level Keras APIs, e.g. self.module.fit(), see [here](https://github.com/tensorflow/tensorflow/issues/21788) ).

I suggest open a [FR/look at where the tf.distribute support goes for custom training loops](https://www.tensorflow.org/beta/guide/distribute_strategy#using_tfdistributestrategy_with_custom_training_loops)  and enforcing single-GPU mode in between

delira/io/tf.py
```diff
@@ -35,3 +37,55 @@ def load_checkpoint(file: str, model=None):
...
+        variable_list += model.variables
+
+    if optimizer is not None:
+        for k, v in optimizer.items():
+            variable_list += v.variables()
+
+    return variable_list
+
+
+def save_checkpoint_eager(file,
```
ORippler, line 56: While this works for now, I suggest switching to [tf.train.Checkpoint](https://www.tensorflow.org/api_docs/python/tf/train/Checkpoint) for both Eager and Graph mode in future.

delira/io/tf.py
```diff
@@ -35,3 +37,55 @@ def load_checkpoint(file: str, model=None):
...
     tf.train.Saver().restore(model._sess, file)
     return {}
+
+
+def _create_varlist(model: AbstractTfEagerNetwork = None,
+                    optimizer: typing.Dict[str, tf.train.Optimizer] = None):
+    variable_list = []
+
+    if model is not None:
+        variable_list += model.variables
```
ORippler, line 47: This raises an actual error instead of a warning due to tf internals if model has not been built or called once.

ValueError: Weights for model sequential have not yet been created. Weights are created when the Model is first called on inputs or `build()` is called with an `input_shape`.

delira/models/backends/tf_eager/abstract_network.py
```diff
@@ -0,0 +1,155 @@
+import abc
+import typing
+import tensorflow as tf
+import numpy as np
+from delira.models.abstract_network import AbstractNetwork
+
+
+class AbstractTfEagerNetwork(AbstractNetwork, tf.keras.layers.Layer):
```
ORippler, line 8: maybe inherit from [tf.Module](https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/Module) in future. Though I might actually prefer tf.keras.layers.Layer, as it provides nice additions/extensions such as trainable flag

delira/training/backends/torch/trainer.py
```diff
@@ -0,0 +1,578 @@
...
+            if latest_state_path is not None:
+
+                # if pth file does not exist, load pt file instead
+                if not os.path.isfile(latest_state_path):
+                    latest_state_path = latest_state_path[:-1]
+
+                logger.info("Attempting to load state from previous \
+                            training from %s" % latest_state_path)
+                try:
+                    self.update_state(latest_state_path)
```
ORippler, line 263: we should set `self.start_epoch = latest_epoch if` we want to resume training. check same for all other trainers.

delira/training/backends/torch/trainer.py
```diff
@@ -0,0 +1,578 @@
...
+        Parameters
+        ----------
+        *args :
+            positional arguments
+        **kwargs :
+            keyword arguments
+
+        """
+        self.save_state(os.path.join(
+            self.save_path, "checkpoint_epoch_0"), 0)
```
ORippler: use self.start_epoch instead of "0"

ORippler (changes requested):
* [x] change https://github.com/delira-dev/delira/blob/cdee72f74a069d2d9919910cf6cf0b6ae96d8373/delira/training/base_trainer.py#L207 to make use of self.start_epoch.
* [x] Look at the docstrings. Oftentimes, e.g. for basetrainer, they are wrong (still wrongly states to raise NotImplementedError. I think this still comes from before the Trainer merger.)

In general, tfeager and tfgraph seem to look quite nice. One note, we don't need to actually do the tf.convert_to_tensor, as tf accepts np.ndarrays. But that is nitpicking.

I also remarked some points for discussion

justusschock:
I'll have a look at all of them and either address them here or in separate PRs:

Regarding the conversation to tensor type: I added this on purpose, since explicit is better than implicit and one can do some optimization (and maybe even quantization) after converting to tensors. Also this shows where the conversion is done, as tensors and numpy arrays are not completely identical regarding their syntax. Also it is important to notice, that for tf eager one will probably start gradient taping right after this. But if you prefer removing this part it's fine for me.

delira/models/backends/tf_eager/data_parallel.py
justusschock: I disabled this for now and will track this in #177

delira/io/tf.py
justusschock: Yeah, but I think it's nicer this way since we don't have to raise an Exception here this way (and this should work in Epochs other than 0)

delira/models/backends/tf_eager/abstract_network.py
justusschock: Added this as reminder to #165

delira/training/backends/torch/trainer.py
justusschock: Done

delira/training/backends/torch/trainer.py
justusschock: I added this, although this was already done by https://github.com/delira-dev/delira/blob/additional_backends/delira/training/backends/torch/trainer.py#L549 

In all other trainers this was already done

delira/io/tf.py
justusschock: tracked in #178

PR #109, opened 2019-05-22
Scikit-Learn Compatibility

justusschock:
This PR aims to add a wrapper for `sklearn`to add compatibility with `delira`.

This is especially useful, if one wants to compare non-deep methods with neural networks (and `sklearn` also offers deep learning modules), because it enables the re-usability of  entire pipeline by simply switching out some classes.

The wrapper is implemented in a way, that it allows training for iterative approaches as neural networks (via `partial_fit`) as well as non-iterative approaches such as KNNs or Decision Trees (via `fit`).

Tests for Sklearn backend are passing, the only tests failing are the tf tests from #66 which aren't relevant for this PR.

This PR should be merged after #66 #98 and #101 have been merged.

Remaining ToDos (will be done, once there are no other changes requested by the reviewers and all previous PRs were merged) :
- [x] Add backend to Setup.py

- [ ] Add backend to README.md

- [ ] Add backend to getting started section of docs

- [ ] Add simple example

PR #110, opened 2019-05-23
Chainer backend

justusschock:
This backend adds a backend for [Chainer](https://chainer.org). `chainer` is quite similar to pytorch and was designed as a python-only DL framework (based on numpy/cupy). Nowadays they are also adding a C++ backend for comparable speed.

This PR should be merged after #66 #98 #101 and #109 have been merged.

Remaining ToDos (will be done, once there are no other changes requested by the reviewers and all previous PRs were merged) :
- [x] Add tests

- [x] Add backend to Setup.py

- [ ] Add backend to README.md

- [ ] Add backend to getting started section of docs

- [ ] Add simple example


delira/training/experiment.py
```diff
@@ -1222,3 +1222,140 @@ def test(self, network, test_data: BaseDataManager,
...
+
+            device = network.device
+            if prepare_batch is None:
+                prepare_batch = partial(network.prepare_batch,
+                                        input_device=device,
+                                        output_device=device)
+
+            # switch to backend-specific convert function
+            if convert_fn is None:
+                convert_fn = convert_chainer_tensor_to_npy
```
mibaumgartner: Why isn't this added as a default argument to the function?

mibaumgartner (approved):
Don't have a lot of `chainer` experience but it looks good to me.

delira/training/experiment.py
justusschock: Will change this

PR #112, opened 2019-05-28
Add stylechecks to CI/CD

justusschock:
Add python stylechecks (first attempt)


justusschock:
Once the checks are running, I will try to update the whole current codebase to correct PEP8 style (maybe also in this issue)

mibaumgartner (approved):

PR #115, opened 2019-05-31
Introduce Debug mode

### LINKED ISSUE #114: [Bug] Debug mode without multiprocessing
**Description**
Unfortunately, Visual Studio Code does not support debugging with `multiprocessing`. A debug mode, where no `multiprocessing` is used, would be beneficial.

justusschock:
This should address and fix #114 

I think currently we only have multiprocessing/threading in dataloading (There is also some threading inside tensorboardX for flushing changes to disk, but we cannot change this one). There might come some additional multiprocessing, once we introduce the new logging scheme, so we should take care of this there too.

I don't know if we additionally want to set debug flags for the backends here?
E.g. we could set `CUDA_LAUCH_BLOCKING=1` which makes it easier to debug cuda code because we get the error message corresponding to the actual failing kernel.

There are also other backend specific flags here, that we might want to set.

Don't know if we want to do this here or if this should be only to disable multiprocessing and the usere should set backend-specific flags himself?

EDIT: This should be merged **after** #66


mibaumgartner (approved):
Works great :D Doc-strings for the context-manager would be great, rest looks good 👍

mibaumgartner:
I think the user should set the debug flags for the respective backend. Otherwise it would be difficult to set all flags for all backends. 
Edit: Furthermore, I think the debug flag should be specific to delira (unitl now, only data loading is  affected) and backend independent.

justusschock:
What do you mean by 
> Furthermore, I think the debug flag should be specific to delira (unitl now, only data loading is affected) and backend independent.

?

mibaumgartner:
When activating the debug flag in delira, only delira functions should be affected by the flag and none of the backends (as it is right now). I think, it would be confusing if other backend specific flags would be triggered in the background.

justusschock:
Thought you wanted to change something :D

PR #133, opened 2019-06-11
Fix predictor caching behavior

### LINKED ISSUE #129: [Bug] Predictor caches results of validation
**Description**
`Predictor` caches the predictions which results in very high memory usage when training segmentation networks and does not fit very well with lazy datasets.

https://github.com/justusschock/delira/blob/9d49560340edd08857262c42b55bd6342c46a0d0/delira/training/predictor.py#L254

justusschock:
Attempts to fix the predictor caching; Disables prediction caching during validation and enables it during testing; 
Fixes #129


mibaumgartner (approved):
Looks good to me :)

PR #134, opened 2019-06-11
Data loading fix

### LINKED ISSUE #127: [Bug] Progress bars display confusing number of batches
**Description**
Progress bars during training show confusing number of batches.
E.g. 
Dataset length: 10 samples
Batch size: 2
results in 8 batches during training and 18 batches during validation even though both sets have the same length (and the numbers are wrong).

**Reproduction**
Should be easy to create a small example.

### LINKED ISSUE #128: [Bug] Debug mode results in wrong behaviour if `num_process_augmentations` is not adjusted
**Description**
When the Delira debug mode is activated but the `num_process_augmentations` inside the `Datamanger` is not adjusted correctly, the number of batches wrong because of the DataLoader and the Sampler.

justusschock:
Fixes #128 
Fixes #127


mibaumgartner (approved):
Looks good to me :)

PR #139, opened 2019-06-12
New logging

justusschock:
First attempt for new logging structure (independent of trixi)


justusschock:
A review on this would be nice, since tests are passing now and we can't proceed in many other things without a new logging ;)

justusschock:
@gedoensmax Tested this and verified it works, 

I'd like to get a review here to move on with other PRs
@mibaumgartner @haarburger

ORippler (changes requested):
In general, I like the new structure. While reviewing, I nonetheless had some questions points open for discussion:

* why do we have many keywords mapping to the same data type? This is a bit confusing, especially when these keywords are tested for in unit_tests (e.g. value and scalar).
* Docstrings are still largely missing .
* Use of Python Multiprocessing requires objects to be pickleable (which is why we need to pass tf.GraphDef instead of tf.Graph).
* It would be nice if we could perform a check in the unit-tests if the objects are actually written/logged. Using TensorboardBackend, I think at least `scalars` and `pr_curve` don't get logged properly, but tests also don't fail. We could assert this by parsing the generated [logfiles](https://stackoverflow.com/questions/37304461/tensorflow-importing-data-from-a-tensorboard-tfevent-file), at least for tensorboard.

justusschock:
Regarding your first point: this came to provide an API with only very few breaking changes from old loggings and to support both, the native visdom and tensorboard keywords, but we can reduce this.

I'll address the second point tomorrow.

Any alternatives for Multiprocessing? I tried threading, but it didn't work properly, and we shouldn't do the logging and computation on the main thread/process.

I don't know how (yet) how to parse the logfiles, but if I got some time, I'll have a look at this or can you make some time for it?

ORippler:
> Regarding your first point: this came to provide an API with only very few breaking changes from old loggings and to support both, the native visdom and tensorboard keywords, but we can reduce this.
> 
> I'll address the second point tomorrow.
> 
> Any alternatives for Multiprocessing? I tried threading, but it didn't work properly, and we shouldn't do the logging and computation on the main thread/process.
> 
> I don't know how (yet) how to parse the logfiles, but if I got some time, I'll have a look at this or can you make some time for it?

I think we should deprecate old logging keys used by trixi and remove them later on in favor of only the visdom and tensorbaord keywords.

:+1: 

Multiprocessing is fine, we should just make the Users aware of the requirement.

I would rather review #102 and afterwards start working on #165 and/or #143 , but I will see if I can make some time.

justusschock:
That's fine, I'll focus on this PR then.

justusschock:
So I added docstrings, deprecated the keys as discussed and the only thing missing here is the check inside the tests

justusschock:
So, I added some basic checks for written tags, there are a few problems though:

* we cannot check for tags always -> For example there is no tag while logging graphs
* With #102 the tests are split between several backends and we can only do this, if `tensorflow` is present, which is why we won't be able to check it for pytorch tensors (and other future tensor types)

justusschock:
So, there are Problems with 2 logging functionalities:

* Logging multiple Scalars at once, which creates separate subdirs with event files
* Logging Text

The issue with multiple scalars could be solved, by simply iterating over the scalar dict and logging them sequentially, which is not that beautiful but should

ORippler:
Docstrings look fine!

Fixed the text logging: in [tensorbaordX](https://github.com/lanpa/tensorboardX/blob/master/tensorboardX/summary.py#L432), the `tag` is altered, so we have to consider that in the unit test. Or you mean sth. else?

Logging multiple values at once I think iteration and sequential logging is fine even if not beautiful. If that problem only applies to TensorboardXWriter, we can also only change it there

With the graphs it's unfortunate but i can't think of an easy proper solution right now.

Same goes for testing.

justusschock:
No, that's exactly what I was talking about.

I changed this in the tensorboard backend and the corresponding tests. 
Anything left to do?

ORippler:
2 Tests for the scalars are seemingly still failing.

Otherwise, except for the removal of trixi and trixi-slim in the requirements it seems good to go!

justusschock:
Yeah, I messed up during test changing. Fixed this now...

Regarding the removal of `trixi` and `trixi-slim`: We cannot do this (yet). We are still building upon some of trixi's modules. Mainly the [`LookupConfig`](https://github.com/delira-dev/delira/blob/new_logging/delira/utils/config.py) is just an extension of `trixi`'s config. But I think, we may just copy that part from them (and give the sublicense there)

EDIT: I'd like to do this in a separate PR, as this one is already huge enough.
If you approve this PR, I'll merge this right after the new bugfix release and work on the new PR to finally remove `trixi`

ORippler (approved):

PR #141, opened 2019-06-13
Remove mutable default arguments

justusschock:
Remove mutable default arguments to avoid unexpected behavior


mibaumgartner (approved):
Looks good :)

PR #143, opened 2019-06-14
WIP: Hyperparam search

justusschock:
This PR adds a basic skeleton for hyperparameter search.

Rather than using [`ray`](https://github.com/ray-project/ray), I started implementing it by myself, because this will make things way easier to integrate.

What's still missing:
- [ ] interation into `experiment` (should be relatively simple) 
- [ ] additional samplers (see [this ray directory](https://github.com/ray-project/ray/tree/master/python/ray/tune/suggest) for examplary samplers)
- [ ] additional schedulers (We may not want to use FIFO only; see [this ray directory](https://github.com/ray-project/ray/tree/master/python/ray/tune/schedulers) for exemplary schedulers).

*Currently trials can only run on a single machine, we might want to change this in future, but this will be a rather large overhead when implementing the basics, so I'm omitting this right now.*

@mibaumgartner @haarburger I'd like to get a first feedback from you regarding the overall structure and any requests, complaints (must not be a complete and detailed review by now). This will probably become a rather huge PR and so early feedback is very welcome, before I implement something, that turns out to be completely useless in the end...


cwern1:
I'll have a look at this over the weekend 👍🏼

PR #145, opened 2019-06-15
Bug fixes

### LINKED ISSUE #144: [Bug] train/test_kwargs emptied during BaseExperiment.kfold
**Description**
During BaseExperiment.kfold, train/test_kwargs are emptied during BaseExperiment.kfold.
Reason is, that [Datamanger.update_state_from_dict](https://github.com/justusschock/delira/blob/master/delira/data_loading/data_manager.py#L406) is called [multiple times](https://github.com/justusschock/delira/blob/master/delira/training/experiment.py#L550).

Inside the function, elements are removed from the input state_dict via .pop. Therefore, train/test_kwargs are emptied during the first update.

Possible solutions: Either pass a copy of train/test_kwargs as input or refrain from altering the input dict inside Datamanager.update_state_from_dict.

justusschock:
* Fixes #144 
* Fixes mutable default arguments during training
* Fixes multiprocessing issues during sampling (moves `sampler` to `Augmenter` instead of `BaseDataLoader`)


delira/training/predictor.py
```diff
@@ -169,7 +169,7 @@ def predict_data_mgr(self, datamgr, batchsize=None, metrics=None,
             a dictionary containing all metrics of the current batch
 
         """
-
+        print(metric_keys)
```
mibaumgartner: Print statement should be removed :)

delira/training/predictor.py
```diff
@@ -282,10 +282,14 @@ def predict_data_mgr_cache_metrics_only(self, datamgr, batchsize=None,
         does not cache anything by default
 
         """
+        print(metric_keys)
```
mibaumgartner: Print state should be removed :)

mibaumgartner (approved):
Please remove the print statement :) Should be ready to merge afterwards.

delira/training/predictor.py
justusschock: I noticed this too and just removed them before your comments :)

justusschock:
@haarburger @ORippler can any of you maybe have a look at the queue logic in `delira.data_loading.datamanager.Augmenter.__next__` and `delira.data_loading.data_loader.BaseDataLoader.generate_train_batch`? This is most likely the part crashing the CI/CD, but run's fine on both of my (and on @mibaumgartner 's) machines, So I cannot debug/reproduce this. 

If I had to guess, the issue lies within the fact, that the `sampler_queue` is empty and thus there's an infinite loop in the `generate_train_batch` function, but usually, this should never happen...

cwern1:
I can do it tomorrow morning.

delira/data_loading/data_manager.py
```diff
@@ -37,6 +40,12 @@ def __init__(self, data_loader: BaseDataLoader, transforms,
         n_process_augmentation : int
             the number of processes to use for augmentation (only necessary if
             not in debug mode)
+        sampler : :class:`AbstractSampler`
+            the sampler to use; must be used here instead of inside the
+            dataloader to avoid duplications and oversampling due to
+            multiprocessing
+        sampler_queues : list of g:class:`multiprocessing.Queue`
```
mibaumgartner: typo

delira/data_loading/data_loader.py
```diff
@@ -22,13 +26,13 @@ def __init__(self, dataset: AbstractDataset,
             dataset to perform sample loading
         batch_size : int
             number of samples per batch
+        sampler_queues : list :class:`multiprocessing.Queue`
```
mibaumgartner: Maybe change to `list of`

delira/data_loading/data_loader.py
```diff
@@ -79,30 +75,33 @@ def generate_train_batch(self):
...
-        else:
-            self._batches_generated += 1
+        idxs = None
+        sampler_queue = self.sampler_queues[self.thread_id]
+        while idxs is None:
+            try:
+                idxs = sampler_queue.get(timeout=0.2)
 
-            idxs = self.sampler(self.batch_size)
+                # idxs = self.sampler_queue.get()
```
mibaumgartner: Do we need this comment?

tmp.py
```diff
@@ -0,0 +1,32 @@
+import numpy as np
+from delira.data_loading import AbstractDataset, BaseDataManager
+
+
+class DummyDataset(AbstractDataset):
```
mibaumgartner: Isn't there a better place and name than to call it `tmp.py`(do we need this file)?

mibaumgartner (approved):
Some small things, rest looks good 👍

tmp.py
justusschock: No we don't :D I already deleted it, this was pushed accidentally an was a debug script for me

PR #152, opened 2019-06-21
add training callbacks

justusschock:


delira/training/callbacks/abstract_callback.py
```diff
@@ -59,3 +59,41 @@ def at_epoch_end(self, trainer, **kwargs):
 
         """
         return {}
+
+    def at_training_begin(self, trainer, **kwargs):
```
mibaumgartner, line 71: I think we should explicitly add the arguments.

delira/training/callbacks/abstract_callback.py
```diff
@@ -59,3 +59,41 @@ def at_epoch_end(self, trainer, **kwargs):
...
+        Returns
+        -------
+        dict
+            modified trainer attributes, where the name must correspond to the
+            trainer's attribute name
+
+        """
+        return {}
+
+    def at_training_end(self, trainer, **kwargs):
```
mibaumgartner, line 90: I think we should explicitly add the arguments.

delira/utils/messenger.py
```diff
@@ -282,7 +282,7 @@ def __init__(self, messenger: BaseMessenger):
         super().__init__()
         self._messenger = messenger
 
-    def at_training_start(self, trainer, **kwargs) -> dict:
+    def at_training_begin(self, trainer, **kwargs) -> dict:
```
mibaumgartner, line 285: Needs to be changed accordingly.

delira/training/base_trainer.py
```diff
@@ -199,8 +199,11 @@ def _at_training_begin(self, *args, **kwargs):
             keyword arguments
 
         """
+
         self.save_state(os.path.join(self.save_path, "checkpoint_epoch_%d"
                                      % self.start_epoch), self.start_epoch)
+        for cbck in self._callbacks:
+            self._update_state(cbck.at_training_begin(self, *args, **kwargs))
```
mibaumgartner, line 203: We should add a test to check if the callbacks are executed as expected.

mibaumgartner (changes requested):
Added some minor comments.

delira/training/callbacks/abstract_callback.py
justusschock: As stated in the other PR ( #172 ) this might be troublesome, since these callbacks are designed to work with every trainer/backend/callback implementation, which might require different arguments here. We would have to discuss what to add and what not to add first

delira/training/callbacks/abstract_callback.py
justusschock: Same as above

mibaumgartner: Closing this for the discussion in the other PR.

delira/training/callbacks/abstract_callback.py
mibaumgartner: Closing this for the discussion in the other PR.

PR #153, opened 2019-06-21
Wrapper for experiments with slack messaging support

mibaumgartner:
Implements a wrapper for experiments to send additional messages about the current status to slack.

- [x] SlackClient
- [x] Docs
- [x] Proper installation
- [x] Tutorial
- [x] Unittest


delira/training/slack.py
```diff
@@ -0,0 +1,311 @@
...
+
+        Returns
+        -------
+        Any
+            attribute
+        """
+        # see if this object has attr
+        # NOTE do not use hasattr, it goes into
+        # infinite recurrsion
+        if attr in self.__dict__:
```
justusschock: Change this to 
```
try:
    return getattr(self, attr)
except AttributeError:
   ...
```

This is the same way `hasattr` works

justusschock (approved):
I requested a minor change.

It is up to you, whether you want to incorporate this change or not, because this is only for code beautification.

I'd recommend you to add docstrings and the tutorial before merging, since the token stuff is non-trivial.

I'd not add this tutorial to the readme, but to the docs.

It's hard to really test this. Is there a way to mock a slack server?

mibaumgartner:
I don't think its possible to mock a slack server but maybe we could check if the corresponding functions are called inside the wrapper? This would at least check the basic functionality of emitting messages.

justusschock:
Seems like a good approach to me although this may break with future slack versions

justusschock:
Something with the dependencies doesn't seem to be right here. I'll have a look tomorrow. I also prefer to postpone merging until we landed #102 since this will change our internal structure quite a bit, so we would have to move things a bit around (I will take care of it once it's time to so so)

justusschock:
And I' currently thinking, whether I would prefer this to be part of the `utils` (or even `logging`) package instead of the `training` package, since it does not implement something for the training process itself like a new trainer or stuff like that

mibaumgartner:
Ok, I think `utils` is a good place 👍

delira/utils/messenger.py
```diff
@@ -0,0 +1,387 @@
...
+from abc import ABC, abstractmethod
+
+from delira.training import BaseExperiment
+from delira.training.callbacks import AbstractCallback
+
+
+class BaseMessenger(ABC):
+    def __init__(self, experiment: BaseExperiment, notify_epochs: int = None):
+        """
+        Wrap arbitrary experiments and connect its functions to a
```
justusschock: Move class description out of init

delira/utils/messenger.py
```diff
@@ -0,0 +1,387 @@
...
+        notify_epochs : int
+            Activates notifications about finished epochs with frequency
+            `notify_epochs`.
+        kwargs :
+            additional keyword arguments passed to :class:`SlackClient`
+
+        See Also
+        --------
+        :class:`BaseMessenger`
+        """
```
justusschock, line 367: There is a `Raises` section missing in this docstring

justusschock (approved):
There are some minor issues with your docstrings, but in general this looks good to me!

PR #154, opened 2019-06-25
New data pipeline

justusschock:
This PR is currently work in progress, but a first review would be nice :)

With this PR we will regain the full control over the dataloading pipeline. This will enable us to easier include bugfixes and changes and adapt it to our framework. Additionally we can more easily provide speedup utilities, like jitting or GPU augmentation. This subpackage is designed with minimal breaking changes for the user (only arguments, which are virtually never used, but were there as a fallback have been changed).

This won't change the batchgenerators dependency, because we still have to use their transforms (it would be much effort and many lines of codeduplication with their repo to completely reimplement them - the potential speedup would be minimal, which currently doesn't make it worth the effort).

The redesign of this subpackage also incorporates the bugfixes from 54dcad96ed6ff4e7830eb1d5303f21676695b492 in a much cleaner and more elegant way.

ToDo:

* [X] Apply Changes from separate (private) working repo to this repo
* [x] Adapt Augmenter integration in trainer 
* [x] Adapt Augmenter integration in predictor
* [x] Adjust tests
* [x] Switch to parallel loading in predictor
* [x] Update notebooks, examples and docs (Still not completely working, but updated with everything that is related to this PR)
* [x] Fix Bug with wrong sample order

> Note: Tests will fail for now, since I did not (yet) have the time to adjust them to the new dataloading pipeline and reintegrate the changes into trainer and predictor.


justusschock:
Remaining ToDo: Fix Sampler Tests

justusschock:
There seems to be a bug regarding the sample order (found by @NKPMedia ), which probably lies within

```
    def __iter__(self):
        # start processes
        self._start_processes()
        _index_pipe = self._next_index_pipe

        # create sampler iterator
        sampler_iter = iter(self._sampler)
        all_sampled = False
```

The proposed solution is to remove the line `_index_pipe = self._next_index_pipe`. WE have to check whether this makes sense and actually resolves the problem

cwern1:
If I run the tests against python 3.5 I get a different error

```bash
======================================================================
ERROR: data_loading.test_augmenters (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: data_loading.test_augmenters
Traceback (most recent call last):
  File "/opt/miniconda3/envs/delira/lib/python3.7/unittest/loader.py", line 436, in _find_test_path
    module = self._get_module_from_name(name)
  File "/opt/miniconda3/envs/delira/lib/python3.7/unittest/loader.py", line 377, in _get_module_from_name
    __import__(name)
  File "/home/staff/haarburger/Workspace/delira/tests/data_loading/test_augmenters.py", line 3, in <module>
    from ..utils import check_for_no_backend
ValueError: attempted relative import beyond top-level package
```

justusschock:
This is not an error, but a wrong usage of our testing environment. You have to execute them via `python -m unittest` to run all tests or `pythin -m test.data_loading.test_augmenters` to only run the tests in this file.

Our tests are written as a separate package using relative imports, which can't be resolved,when you don't execute it as a module but as a file

cwern1:
Ok, the test gets stuck at
```bash
(delira-py35) haarburger@pc50:~/Workspace/delira$ python -m unittest --verbose
/opt/miniconda3/envs/delira-py35/lib/python3.5/site-packages/tensorboard/compat/tensorflow_stub/dtypes.py:541: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_qint8 = np.dtype([("qint8", np.int8, 1)])
/opt/miniconda3/envs/delira-py35/lib/python3.5/site-packages/tensorboard/compat/tensorflow_stub/dtypes.py:542: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_quint8 = np.dtype([("quint8", np.uint8, 1)])
/opt/miniconda3/envs/delira-py35/lib/python3.5/site-packages/tensorboard/compat/tensorflow_stub/dtypes.py:543: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_qint16 = np.dtype([("qint16", np.int16, 1)])
/opt/miniconda3/envs/delira-py35/lib/python3.5/site-packages/tensorboard/compat/tensorflow_stub/dtypes.py:544: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_quint16 = np.dtype([("quint16", np.uint16, 1)])
/opt/miniconda3/envs/delira-py35/lib/python3.5/site-packages/tensorboard/compat/tensorflow_stub/dtypes.py:545: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_qint32 = np.dtype([("qint32", np.int32, 1)])
/opt/miniconda3/envs/delira-py35/lib/python3.5/site-packages/tensorboard/compat/tensorflow_stub/dtypes.py:550: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  np_resource = np.dtype([("resource", np.ubyte, 1)])
/opt/miniconda3/envs/delira-py35/lib/python3.5/site-packages/tensorflow/python/framework/dtypes.py:516: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_qint8 = np.dtype([("qint8", np.int8, 1)])
/opt/miniconda3/envs/delira-py35/lib/python3.5/site-packages/tensorflow/python/framework/dtypes.py:517: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_quint8 = np.dtype([("quint8", np.uint8, 1)])
/opt/miniconda3/envs/delira-py35/lib/python3.5/site-packages/tensorflow/python/framework/dtypes.py:518: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_qint16 = np.dtype([("qint16", np.int16, 1)])
/opt/miniconda3/envs/delira-py35/lib/python3.5/site-packages/tensorflow/python/framework/dtypes.py:519: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_quint16 = np.dtype([("quint16", np.uint16, 1)])
/opt/miniconda3/envs/delira-py35/lib/python3.5/site-packages/tensorflow/python/framework/dtypes.py:520: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_qint32 = np.dtype([("qint32", np.int32, 1)])
/opt/miniconda3/envs/delira-py35/lib/python3.5/site-packages/tensorflow/python/framework/dtypes.py:525: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  np_resource = np.dtype([("resource", np.ubyte, 1)])
test_parallel (tests.data_loading.test_augmenters.TestAugmenters) ...
```

justusschock:
Do you have any idea why? I will try to reproduce this, but this was not possible on my machine so far

cwern1:
I have no idea to be honest but I will try to dig deeper with the debugger.

cwern1:
My debugger keeps crashing if I try to dig deeper. Since this error only occurs on some systems and on older Python version, should we just ignore it for now? I think at least @NKPmedia, @mibaumgartner and I have already run several evaluations on this branch without experiencing this issue here...

justusschock:
I must admit, that I don't like this approach as it would completely break our python 3.5 and python 3.6 tests and I don't want anything on master, we don't understand. This might have unexpected behavior in future changes

cwern1:
I think we misunderstood each other here: I didn't suggest to merge this with failing tests but rather drop support for Python 3.5 and 3.6.

justusschock:
I think this would limit our userbase quite hard as there are still many libraries written in Python 3.5 which are not compatible with newer versions.

cwern1:
Do you have a few examples? As far as I can tell most current users are working with Python 3.7...

delira/data_loading/dataset.py
```diff
@@ -211,6 +212,36 @@ def __next__(self):
...
+        return self[index]
+
+    def _make_dataset(self, path: str):
+        pass
+
+
+class IterableDataset(AbstractDataset):
+    def __init__(self, data: Iterable):
+        super().__init__(None, None)
+        self._data = data
```
mibaumgartner: Please add a test and docstring for this

delira/data_loading/dataset.py
```diff
@@ -211,6 +212,36 @@ def __next__(self):
         return sample
 
 
+class DictDataset(AbstractDataset):
+    def __init__(self, data: dict):
+        super().__init__(None, None)
+        self._data = data
```
mibaumgartner: Please add a test and docstring for this

delira/data_loading/data_loader.py
```diff
@@ -1,118 +1,101 @@
...
-                    result_dict[key] = np.asarray(val_list)
+        return data_dict
 
-                return result_dict
-            except Empty:
-                pass
-
-    def _get_sample(self, index):
+    @property
+    def process_id(self):
```
mibaumgartner, line 68: Please add a test for the property methods

delira/data_loading/data_loader.py
```diff
@@ -1,118 +1,101 @@
...
-        self.sampler_queues = sampler_queues
+        else:
+            # wrap it into dataset depending on datatype
+            if isinstance(data, dict):
+                dataset = DictDataset(data)
+            elif isinstance(data, Iterable):
+                dataset = IterableDataset(data)
+            else:
+                raise TypeError("Invalid dataset type: %s"
+                                % type(data).__name__)
```
mibaumgartner, line 34: The tests do not cover these cases

delira/data_loading/data_loader.py
```diff
@@ -1,118 +1,101 @@
...
+        data : Any
+            the data to use; Ideally this either is a dataset, an iterable or
+            a dict, but in general, this must only be indexable, have a length
+            and return a dict of arrays if indexed
         """
+        self._process_id = None
 
-        # store dataset in self._data
-        super().__init__(dataset, batch_size)
+        # do nothing if data
```
mibaumgartner: Nit: the comment might be unnecessary

delira/data_loading/data_loader.py
```diff
@@ -1,118 +1,101 @@
...
+        data_dict = {}
 
-                result_dict = {}
+        # concatenate dict entities by keys
+        for _result_dict in data:
+            for key, val in _result_dict.items():
+                if key in data_dict.keys():
+                    data_dict[key].append(val)
+                else:
+                    data_dict[key] = [val]
```
mibaumgartner: Nit: eventually replace with defaultdict (it is initialized right above, so the meaning is not very hidden)

delira/data_loading/sampler/abstract.py
```diff
@@ -0,0 +1,56 @@
...
+        """
+        Returns an iterator, must be overwritten in subclasses
+
+        Raises
+        ------
+        NotImplementedError
+            if not overwritten in subclass
+
+        """
+        raise NotImplementedError
```
mibaumgartner, line 29: Nit: eventually add test to increase coverage in the future

delira/data_loading/sampler/batch.py
```diff
@@ -0,0 +1,65 @@
+from delira.data_loading.sampler.abstract import AbstractSampler
+
+
+class BatchSampler(object):
```
mibaumgartner, line 4: I'm not sure about the purpose of this class. It does something similar as the `DataLoader` without resolving the internal dicts. What is the usecase of this sampler? (Could we replace the `DataLoader` with this sampelr?)

delira/data_loading/sampler/batch.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self, sampler: AbstractSampler, batch_size, truncate=False):
+        """
+
+        Parameters
+        ----------
+        sampler : :class:`AbstractSampler`
+            the actual sampler producing single-sized samples
+        batch_size : int
+            the size of each batch
+        truncate : bool
```
mibaumgartner: Nit: use consistent naming of this argument (in `DataManager` it is called `drop_last ` which i prefer)

delira/data_loading/data_manager.py
```diff
@@ -340,38 +78,30 @@ class defining the sampling strategy
...
+        if not isinstance(data, AbstractDataset):
+            if isinstance(data, dict):
+                data = DictDataset(data)
+            elif isinstance(data, Iterable):
+                data = IterableDataset(data)
             else:
-                assert issubclass(dataset_cls, AbstractDataset), \
-                    "dataset_cls must be subclass of AbstractDataset"
+                raise TypeError("Invalid Data type given: %s"
+                                % type(data).__name__)
```
mibaumgartner: Duplicate to `DataLoader`. I would prefer adding a function for this or it should be handled at one place.
(And a test is missing for this ^^)

delira/data_loading/data_manager.py
```diff
@@ -437,13 +160,10 @@ def get_subset(self, indices):
             "transforms": self.transforms,
             "sampler_cls": self.sampler.__class__,
             "data_loader_cls": self.data_loader_cls,
```
mibaumgartner, line 162: `drop_last ` and `sample_kwargs` are missing

delira/data_loading/data_manager.py
```diff
@@ -483,55 +203,17 @@ def update_state_from_dict(self, new_state: dict):
         self.data_loader_cls = new_state.pop("data_loader_cls",
                                              self.data_loader_cls)
         # update
-        new_sampler = new_state.pop("sampler", None)
+        new_sampler = new_state.pop("sampler_old", None)
```
mibaumgartner: Why is this `sampler_old`?

delira/data_loading/data_manager.py
```diff
@@ -483,55 +203,17 @@ def update_state_from_dict(self, new_state: dict):
...
         if new_sampler is not None:
             self.sampler = new_sampler.from_dataset(
-                self.dataset,
+                self.data,
                 **new_state.pop("sampling_kwargs", {}))
```
mibaumgartner: I think, by popping sampler kwargs first, the creation of the sampler could be handled by the property setter

delira/data_loading/data_manager.py
```diff
@@ -762,6 +416,29 @@ def n_batches(self):
...
 
         return n_batches
+
+    @property
+    def dataset(self):
+        return self.data
+
+    @dataset.setter
+    def dataset(self, new_dset):
+        assert isinstance(new_dset, AbstractDataset)
```
mibaumgartner: Nit: I prefer `TypeErrors` instead of assertion errors, because this actually checks input arguments from the user.
(applies to all checks in this file, `__init__` raises `TypeError` and assertions which is inconsistent)
Furthermore, I think other libraries raise `TypeError` for wrong arguments.

delira/data_loading/augmenter.py
```diff
@@ -0,0 +1,520 @@
...
+import sys
+import numpy as np
+import random
+
+from delira.data_loading.sampler import AbstractSampler, BatchSampler
+from delira.data_loading.data_loader import DataLoader
+from delira import get_current_debug_mode
+
+
+class _WorkerProcess(multiprocessing.Process):
```
mibaumgartner: Nit: I would prefer moving this behind `_ParallelAugmenter` where it is used and not at the beginning of this file.

delira/data_loading/augmenter.py
```diff
@@ -0,0 +1,520 @@
...
+
+            # create the actual process
+            process = _WorkerProcess(dataloader=self._data_loader,
+                                     output_pipe=send_conn_out,
+                                     index_pipe=recv_conn_in,
+                                     transforms=self._transforms,
+                                     abort_event=self._abort_event,
+                                     process_id=i)
+
+            # make the process daemonic and start it
```
mibaumgartner: Nit: Comment might be unnecessary,

delira/data_loading/augmenter.py
```diff
@@ -0,0 +1,520 @@
...
+
+    def _shutdown_processes(self):
+        """
+        Shuts down the processes and resets all related flags and counters
+        """
+
+        # create copy to avoid modifying the list we iterate over
+        worker = list(
+            zip(self._data_pipes, self._index_pipes, self._processes))
+        # for each process:
```
mibaumgartner: Nit: Comment might be unnecessary,

delira/data_loading/augmenter.py
```diff
@@ -0,0 +1,520 @@
...
+        for _data_conn, _index_conn, _process in worker:
+
+            # send None to worker
+            _index_conn.send(None)
+            # Close Process and wait for its termination
```
mibaumgartner: Nit: Comment might be unnecessary,

delira/data_loading/augmenter.py
```diff
@@ -0,0 +1,520 @@
...
+        Shuts down the processes and resets all related flags and counters
+        """
+
+        # create copy to avoid modifying the list we iterate over
+        worker = list(
+            zip(self._data_pipes, self._index_pipes, self._processes))
+        # for each process:
+        for _data_conn, _index_conn, _process in worker:
+
+            # send None to worker
```
mibaumgartner: Nit: Comment might be unnecessary,

delira/data_loading/augmenter.py
```diff
@@ -0,0 +1,520 @@
...
+            if sys.version_info >= (3, 7):
+                _process.close()
+            else:
+                _process.terminate()
+
+            # close connections
+            _index_conn.close()
+            _data_conn.close()
+
+            # pop corresponing pipes, counters and processes from lists
```
mibaumgartner: Nit: Comment might be unnecessary,

delira/data_loading/augmenter.py
```diff
@@ -0,0 +1,520 @@
...
+            # send None to worker
+            _index_conn.send(None)
+            # Close Process and wait for its termination
+            _process.join()
+            if sys.version_info >= (3, 7):
+                _process.close()
+            else:
+                _process.terminate()
+
+            # close connections
```
mibaumgartner: Nit: Comment might be unnecessary,

delira/data_loading/augmenter.py
```diff
@@ -0,0 +1,520 @@
...
+
+        # receive data from worker
+        data = self._data_pipes[_data_pipe].recv()
+        # decrease number of enqueued batches for current worker
+        self._data_queued[_data_pipe] -= 1
+
+        return data
+
+    def __iter__(self):
+        # start processes
```
mibaumgartner: Nit: Comment might be unnecessary,

delira/data_loading/augmenter.py
```diff
@@ -0,0 +1,520 @@
...
+        self._start_processes()
+
+        # create sampler_old iterator
```
mibaumgartner: Nit: Comment might be unnecessary, 
What is `sampler_old` iterator and where is the difference to a "new" sampler?

delira/data_loading/augmenter.py
```diff
@@ -0,0 +1,520 @@
...
+                else:
+                    break
+
+        except Exception as e:
+            # set abort event to shutdown workers
+            self._abort_event.set()
+            raise e
+
+        finally:
+            # shutdown processes
```
mibaumgartner: Nit: Comment might be unnecessary,

delira/data_loading/augmenter.py
```diff
@@ -0,0 +1,520 @@
...
+            except StopIteration:
+                all_sampled = True
+
+            self._enqueue_indices(_indices)
+
+            # iterate while not all data has been sampled and any data is
+            # enqueued
+            while True:
+
+                # break if abort event has been set
```
mibaumgartner: Nit: Comment might be unnecessary,

delira/data_loading/augmenter.py
```diff
@@ -0,0 +1,520 @@
...
+            the basic seed; default: 1
+        drop_last : bool
+            whether to drop the last (possibly smaller) batch or not
+        """
+        super().__init__(data_loader=data_loader, batchsize=batchsize,
+                         sampler=sampler, transforms=transforms, seed=seed,
+                         drop_last=drop_last)
+
+    def __iter__(self):
+        # create sampler_old iterator
```
mibaumgartner, line 426: Nit: Comment might be unnecessary,

delira/data_loading/augmenter.py
```diff
@@ -0,0 +1,520 @@
...
+        """
+
+        self._data_loader = data_loader
+
+        if not isinstance(sampler, BatchSampler):
+            if isinstance(sampler, AbstractSampler):
+                sampler = BatchSampler(sampler, batchsize,
+                                       truncate=drop_last)
+            else:
+                raise ValueError("Invalid Sampler given: %s" % str(sampler))
```
mibaumgartner: Change to `TypeError` for consistency

delira/data_loading/sampler/random.py
```diff
@@ -0,0 +1,101 @@
...
+        Parameters
+        ----------
+        indices : list
+            the indices containing the classes to sample from
+        num_samples : int
+            number of samples to provide, if not specified: defaults to the
+            amount values given in :param:`indices`
+
+        """
+        super().__init__(indices, True, num_samples)
```
mibaumgartner, line 101: Do we really need these?

delira/data_loading/sampler/random.py
```diff
@@ -0,0 +1,101 @@
...
+    def __len__(self):
+        """
+        Defines the length of the sampler
+
+        Returns
+        -------
+        int
+            the number of samples
+        """
+        if self._replacement:
```
mibaumgartner, line 61: There might be a test missing for the length

delira/data_loading/sampler/weighted.py
```diff
@@ -0,0 +1,95 @@
...
+    def __len__(self):
+        """
+        Defines the length of the sampler
+
+        Returns
+        -------
+        int
+            the number of samples
+        """
+        return self._num_samples
```
mibaumgartner, line 50: There might be a test missing for the length

tests/data_loading/test_augmenters.py
```diff
@@ -0,0 +1,81 @@
...
+        dataset = DummyDataset(self._dset_len)
+        data_loader = DataLoader(dataset)
+        sampler = SequentialSampler.from_dataset(dataset)
+
+        if "parallel" in self._testMethodName:
+            self.aug = Augmenter(data_loader, self._batchsize, sampler, 2,
+                                 drop_last=self._drop_last)
+        else:
+            self.aug = Augmenter(data_loader, self._batchsize, sampler, 0,
+                                 drop_last=self._drop_last)
```
mibaumgartner, line 29: I would prefer to create a function which can be called from the respective test where the arguments `parallel` and `drop_last` can be set.

tests/data_loading/test_augmenters.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @unittest.skipUnless(check_for_no_backend(),
+                         "Test should be only executed if no "
+                         "backend was installed")
+    def test_sequential_drop_last(self):
+        self._aug_test()
+
+
+if __name__ == '__main__':
+    unittest.main()
```
mibaumgartner, line 126: Can we add a test where transforms are called? I think this could be easily done with a mock object

tests/data_loading/test_data_manager.py
```diff
@@ -33,10 +34,12 @@ def test_base_datamanager(self):
...
 
-        for key, val in next(manager.get_batchgen()).items():
+        augmenter = manager.get_batchgen()
+        augmenter_iter = iter(augmenter)
+        for key, val in next(augmenter_iter).items():
             self.assertTrue((val == batch_dict[key]).all())
 
-        for key, val in next(manager.get_batchgen()).items():
+        for key, val in next(augmenter_iter).items():
             self.assertEqual(len(val), batch_size)
```
mibaumgartner, line 45: I would like to propose to add some additional integration tests here.

2 Tests: Check with mock object if transforms are called correctly. (serial and parallel)

Additional Tests: Create a dataset with an internal ordering (something like an ID) and check if the sampling order is correct. This should be done with a sequential sampler and a random sampler (serial and parallel). The random sampler should be tested with a seed and check the first few sampler + sampling of all indices (drop_last should be false).

tests/data_loading/test_sampler.py
```diff
@@ -1,174 +1,102 @@
-import unittest
+# ToDo: Add sampler tests
```
mibaumgartner: TODO comment

tests/data_loading/test_sampler.py
```diff
@@ -1,174 +1,102 @@
...
+    RandomSamplerNoReplacement, BatchSampler
 
 from ..utils import check_for_no_backend
-
 from .utils import DummyDataset
 
 
 class SamplerTest(unittest.TestCase):
+    def setUp(self) -> None:
+        self.dset = DummyDataset(600, [0.5, 0.3, 0.2])
```
mibaumgartner, line 13: I would like a different dataset with an internal ordering with ids, so we can check correct sampling of them.
sequential: sequential order
random: check a few samples with seed + check if all samples are sampled 
prevalence: check for average present labels for large number of samples with some boundaries

mibaumgartner (changes requested):
Other questions:
1. https://github.com/delira-dev/delira/blob/499b72db0d042e6cc6f43fdaf2f87fa19a792e89/delira/data_loading/data_manager.py#L256-L258 `Augmenter` uses the `_SequentialAugmenter` when processes is set to 0, not 1. This should be consistent between them. 

Added some comments :)

mibaumgartner:
With regard to the importance of the PR, it might make sense to summarise some comments into an Issue which is handled separately.

delira/data_loading/sampler/batch.py
justusschock: No we can't. This sampler builds batches of indices (The sampler classes only return single indices now). The DataLoader mainly takes these indices, loads the corresponding data and combines them to single arrays.

The pipeline is as follows:
A sampler is created and wrapped by this class (to get batches of indices). It lives in the main process and sends these batches to the worker processes. These processes now load the data, combine them in batches and send them back to the main process.

We could instead include the batch-building in the samplers, but I really prefer it this way, since including the logic of dropping the last batch (and especially the checks for that) complicate the sampling algorithms beyond necessity.

delira/data_loading/data_manager.py
justusschock: I removed this here in favor of dynamically creating the sampler after the dataloader

delira/data_loading/data_manager.py
justusschock: It isn't any longer

delira/data_loading/data_manager.py
justusschock: This is obsolete by dynamically creating the sampler

delira/data_loading/augmenter.py
justusschock: This is not up to date anymore. This was part of the intermediate state of this PR, simply forgot to change this

delira/data_loading/sampler/random.py
justusschock: We don't need them, they are just for the user's convenience. Opinions @haarburger @ORippler ?

delira/data_loading/sampler/weighted.py
justusschock: I don't think such a test is necessary here. We only return a single fixed value here

tests/data_loading/test_augmenters.py
justusschock: This somehow is the function you're asking for :D

tests/data_loading/test_augmenters.py
justusschock: To be honest, I'm not sure, how a mock object would help here...

tests/data_loading/test_data_manager.py
justusschock: I added tests for sampling order, but IMO the tests with random sampler and indices do not belong here, but to the sampler tests (if anywhere). But since we use well tested libraries for sampling, I don't think we need them at all.

tests/data_loading/test_sampler.py
justusschock: This is the only thing I would prefer to move to a separate issue.

tests/data_loading/test_data_manager.py
mibaumgartner: I thought one of the main problems solved in this PR is the correct sampling with the parallel augmenter (the whole purpose of the pipe construct). Therefore, there should be a test where the correctness of this is assessed.

tests/data_loading/test_augmenters.py
mibaumgartner: You could pass a mock object to the augmenter and check if it was called.

tests/data_loading/test_augmenters.py
mibaumgartner: I thought it would be nice to have it as a separate function with arguments and not by checking the function name ..

delira/data_loading/sampler/weighted.py
mibaumgartner: I think, our goal should be 100% coverage. This also includes testing these small things. Other opinions @haarburger @ORippler ?

tests/data_loading/test_augmenters.py
justusschock: Then I would propose not to do it in this PR (mainly because we are doing it the same way in other tests), but to do it during a test-refactoring.

tests/data_loading/test_data_manager.py
justusschock: We had correct sampling before. The main problem solved here is the speed. But I agree that in the long run, we should have a test for this. Are you ok with postponing this in a separate Issue too or do you want to have it included here?

mibaumgartner: We can add it to a separate Issue, just wanted to mention it for the future :)

tests/data_loading/test_augmenters.py
justusschock: Unfortunately this is not possible, because the Mock objects aren't pickleable and therefore they don't work with multiprocessing. The approach to do so with logging was also not successful, since the logging of daemonic processes is not send to the main process

requirements/base.txt
```diff
@@ -10,6 +10,7 @@ pylint
 tqdm
 visdom>=0.1.8.5
 pyyaml
+trixi-slim>0.1.2.1
```
mibaumgartner: I think we don't need trixi anymore.

requirements/tensorflow.txt
```diff
@@ -1 +1,3 @@
 tensorflow-gpu==1.14
+# use trixi-slim >= 0.1.2.1 for tensorboardX support
+trixi-slim>0.1.2.1
```
mibaumgartner: I think we don't need trixi anymore.

requirements/torch.txt
```diff
@@ -1,2 +1,3 @@
 torchvision>=0.2.1
 torch>=1.0.0
+trixi>0.1.2.1
```
mibaumgartner: same as above.

mibaumgartner (approved):
I think the requirements aren't correct, the rest looks good to me.

justusschock:
Yeah, the requirements snook in again during the merge of `master` branch I guess

PR #161, opened 2019-07-09
Add Versioneer for versioning

justusschock:
This will enable automatic versioning for our package and additionally provide unique version names based on the latest tag and the last commit.

I want to integrate this, since it makes debugging and comparing versions much simpler.


mibaumgartner (approved):
Looks good 👍

PR #162, opened 2019-07-09
Refactor __init__.py

justusschock:
Refactors the global `__init__.py` to clean internal scopes


delira/_backends.py
```diff
@@ -0,0 +1,73 @@
...
+# each backend should consist of a tuple of length 2 with the first entry
+# being the package import name and the second being the backend abbreviation.
+# E.g. TensorFlow's package is named 'tensorflow' but if the package is found,
+# it will be considered as 'tf' later on
+__POSSIBLE_BACKENDS = [("torch", "torch"), ("tensorflow", "tf")]
+__BACKENDS = []
+
+
+def _determine_backends():
+
```
mibaumgartner, line 21: Docstring please :)

delira/_backends.py
```diff
@@ -0,0 +1,73 @@
...
+                      f, sort_keys=True, indent=4)
+
+        del _backends
+
+    # set values from config file to variable
+    with open(_config_file) as f:
+        _config_dict = json.load(f)
+    for key, val in _config_dict.pop("backend").items():
+        if val:
+            __BACKENDS.append(key.upper())
```
mibaumgartner, line 63: Any reason why it is `upper()` and not `lower()`? I would prefer `lower()` to be consistent with `__POSSIBLE_BACKENDS ` at the top.

mibaumgartner (approved):
Looks good to me :) Some minor questions but nothing serious.

delira/_backends.py
justusschock: It has been upper, because this somehow are constants we are checking in the rest of the code, and by default, constants should be uppercase. What do the others think on this? @haarburger @ORippler 

Also this is the same as done before :) It's just some refactoring, no breaking changes

delira/_backends.py
justusschock: Done!

delira/_backends.py
justusschock: So I'll merge this now with uppercase, if we want to change this, we can still do this somewhere else

PR #166, opened 2019-07-26
Remove trixi

justusschock:
This PR aims to remove trixi from our requirements. The only thing from `trixi` we need, is a part of their utils module: the [`Config`](https://github.com/delira-dev/delira/blob/remove_trixi/delira/utils/_external/trixi/config.py) and their [`Encoders/Decoders`](https://github.com/delira-dev/delira/blob/remove_trixi/delira/utils/_external/trixi/json_coding.py). This is now inside a separate [package](https://github.com/delira-dev/delira/tree/remove_trixi/delira/utils/_external) with a corresponding [sublicense](https://github.com/delira-dev/delira/blob/remove_trixi/delira/utils/_external/trixi/LICENSE).

The import of the old logging Handlers will work if `trixi` is installed, but raise a deprecation warning, the tests for these handlers can also only be run if `trixi` is installed and will be skipped 

Neither `trixi` nor `trixi-slim` will be installed as a requirement anymore.

This is a follow-up PR to #139


justusschock:
This can be only tested after #139 was merged

delira/utils/config.py
```diff
@@ -1,5 +1,142 @@
...
+    # TODO: support for json, yaml and argparse
+    # TODO: support for loading complex objects
+    def load():
+        raise NotImplementedError
+
+    def loads():
+        raise NotImplementedError
+
+    # TODO: check if copy and deepcopy works out of the box
+    def __copy__(self, *args, **kwargs):
```
justusschock: This probably won't work, since it says for me, that `dict` has no such attribute, maybe leave this one out or implement it from scratch

delira/utils/config.py
```diff
@@ -1,5 +1,142 @@
...
+        super().__copy__(*args, **kwargs)
+
+    def __deepcopy__(self, *args, **kwargs):
```
justusschock: same as for `__copy__`

PR #168, opened 2019-07-27
Add Test-Time-Augmentation to Predictor

justusschock:
This aims to provide optional Test-Time Augmentation.

We only need to add the arguments to the Trainers if we want to keep it that way.


delira/training/predictor.py
```diff
@@ -27,7 +28,9 @@ class Predictor(object):
     def __init__(
             self, model, key_mapping: dict,
             convert_batch_to_npy_fn=convert_batch_to_numpy_identity,
-            prepare_batch_fn=lambda x: x, **kwargs):
+            prepare_batch_fn=lambda x: x,
+            tta_transforms: tuple = (), tta_reduce_fn=None,
+            tta_transform_back=False, **kwargs):
```
mibaumgartner: I would prefer a second tuple with corresponding inverse transformations (instead of an additional method inside of each transofrmation). I think, it would improve compatibility with Batchgenerators.

delira/training/predictor.py
```diff
@@ -553,3 +606,61 @@ def calc_metrics(batch: LookupConfig, metrics=None, metric_keys=None):
...
+                results = []
+                for trafo in transforms:
+                    # apply transforms if possible
+                    if trafo is not None:
+                        data_dict = trafo(**data_dict)
+
+                    _result = function(data_dict, **kwargs)
+
+                    if transform_back:
+                        _result = trafo.inverse(**_result)
```
mibaumgartner: Add an additional condition to check if the `trafo` is not `None`. (I think otherwise this would crash)

delira/training/predictor.py
```diff
@@ -27,7 +28,9 @@ class Predictor(object):
     def __init__(
```
mibaumgartner: I'm not sure if we should keep the original predictor and add this as a subclass.

mibaumgartner (changes requested):
Please add an additional test with tta.
I like the idea of the decorator 👍

delira/training/predictor.py
justusschock: We could do this instead of  the `tta_transform_back` argument and only transform back if trafos are passed here

delira/training/predictor.py
justusschock: I would not like to add this to a subclass, since we would have to re-derive all trainers if we intend to use TTA during training. This is why I originally came up with the idea of the decorator, since this allows us to maintain the current default behavior.

justusschock:
> Please add an additional test with tta.

I will.

> I like the idea of the decorator 👍

Do you know an elegant method to speed up the transformations in a parallel way? we shouldn't do multiprocessing here due to possible problems regarding the async behavior and initialization of CUDA-Streams. Probably some sort of jitting would be the best way here (e.g. via `numba`). 

Do we want to integrate this per default, if numba is installed? Or should this be done by the user?

mibaumgartner:
Before we integrate some kind of jitting for speed up we should investigate if jitting is actually worth it.

justusschock:
So, I added the inverse transforms as a explicit argument, but we have to decide, whether we want to keep it in the base predictor before writing additional tests

delira/training/predictor.py
```diff
@@ -553,3 +609,83 @@ def calc_metrics(batch: LookupConfig, metrics=None, metric_keys=None):
...
+                        if _inv_trafo is not None:
+                            _result = _inv_trafo(**_result)
+
+                    # add each item in current result dict to total results
+                    # dict
+                    for k, v in flatten_dict(_result).items():
+                        if k in results:
+                            results[k] += [v]
+                        else:
+                            results[k] = [v]
```
mibaumgartner, line 789: Maybe make `result` to a defualtdict with list and just use the `append` function. With this the if statement is not needed anymore.

delira/training/predictor.py
```diff
@@ -553,3 +609,83 @@ def calc_metrics(batch: LookupConfig, metrics=None, metric_keys=None):
...
+                    # check whether to transform back
+                    if transform_back:
+                        _inv_trafo = inverse_transforms[idx]
+                        # check if trafo must be applied
+                        if _inv_trafo is not None:
+                            _result = _inv_trafo(**_result)
+
+                    # add each item in current result dict to total results
+                    # dict
+                    for k, v in flatten_dict(_result).items():
```
mibaumgartner, line 785: Why is it neccessary to flatten `_result`?

delira/training/predictor.py
```diff
@@ -553,3 +609,83 @@ def calc_metrics(batch: LookupConfig, metrics=None, metric_keys=None):
...
+                        if k in results:
+                            results[k] += [v]
+                        else:
+                            results[k] = [v]
+
+                # convert every value list inside the results dict to arrays
+                # and apply the reduce function afterwards.
+                for k, v in results.items():
+                    results[k] = reduce_fn(np.array(v))
```
mibaumgartner, line 794: I'm not sure if I would prefer the whole dict as the input to the `reduce_fn` because definitely exist cases where the function/the same function should be applied to the whole dict. E.g. when using an Identifier for the data. But I'm not sure if those cases are too special and may overcomplicate things.

tests/training/test_predictor.py
```diff
@@ -0,0 +1,37 @@
...
+
+        self._predictor = Predictor(model=network_dummy_fn,
+                                    key_mapping={"x": "data"},
+                                    tta_transforms=(
+                                        RangeTransform(),
+                                        ContrastAugmentationTransform()
+                                    ),
+                                    tta_reduce_fn=dummy_reduce_fn)
+
+        preds = self._predictor({"data": np.random.rand(4, 3, 32, 32)})
```
mibaumgartner, line 43: Please extend the test with an inverse transformation (this covers more functionality and improves coverage a little bit more).

tests/training/test_predictor.py
```diff
@@ -0,0 +1,37 @@
+from tests.utils import check_for_no_backend
```
mibaumgartner: Didn't work locally on may machine. The other tests are using a relative import, maybe change this to be consistent.

mibaumgartner (changes requested):
Some minor personal preferences, rest looks good.

delira/training/predictor.py
justusschock: Because TF Graph Backend returns nested dicts including all losses and this is the easiest way to deal with it, without other explicit recursions

delira/training/predictor.py
justusschock: Well this wouldn't be faster and less implicit, so I would prefer the current way

mibaumgartner:
`if __name__ == "__main__":  unittest.main()` is missing in the test file 🙈

delira/training/predictor.py
justusschock: I think this would be to complicated in this case, since this is a good standard value and usually there is only one prediction, that should be reduced. Also one could do this via the transforms

PR #171, opened 2019-07-30
Add seed_all function to seed all backends

### LINKED ISSUE #82: [FeatureRequest]Seed-all fn
I'd like to add a `seed_all` fn to the basic `delira` namespace, which seeds numpy nd all available backends (and maybe python's `random` just to be sure?). 

what do you think @haarburger  @ORippler ?

justusschock:
Closes #82


delira/_backends.py
```diff
@@ -83,3 +83,37 @@ def get_backends():
...
+
+    import numpy as np
+    np.random.seed(seed)
+
+    import random
+    random.seed = seed
+
+    if "torch" in sys.modules and "TORCH" in get_backends():
+        import torch
+        torch.random.manual_seed(seed)
```
mibaumgartner, line 111: https://pytorch.org/docs/stable/notes/randomness.html uses `torch.manual_seed`, is there a difference?

mibaumgartner (approved):
Minor question, rest looks good to go :)

delira/_backends.py
justusschock: No it isn't: https://github.com/pytorch/pytorch/blob/master/torch/__init__.py#L180

PR #172, opened 2019-07-31
New metric calculation

justusschock:
This PR aims to unify metric calculation for testing and training. This is done by moving the metric calculation outside the `closure`.

*This PR includes breaking changes* 

This PR should be merged once #102 is merged

Since the topics are related, I integrated the PR #175 in here.


tests/utils/test_messenger.py
```diff
@@ -46,24 +46,28 @@ class DummyTrainer(BaseNetworkTrainer):
...
-            self._at_epoch_end(None, epoch, is_best)
+            self._at_epoch_end({}, None, epoch, is_best)
         self._at_training_end()
         return DummyNetwork()
 
     def test(self, *args, **kwargs):
         return [{}], [{}]
 
+    def save_state(self, file_name, *args, **kwargs):
+        pass
```
cwern1, line 69: Is there anything missing here?

delira/logging/base_logger.py
```diff
@@ -2,6 +2,179 @@
 from queue import Full
 from delira.logging.base_backend import BaseBackend
 import logging
+import numpy as np
+from types import FunctionType
+from collections import Iterable, MutableMapping
+
+
+# Reduction Functions
+def _reduce_last(items: list):
```
mibaumgartner: Applies to alle reduction functions: add return type.

delira/logging/base_logger.py
```diff
@@ -2,6 +2,179 @@
...
+    Reduction Function returning the last element
+
+    Parameters
+    ----------
+    items : list
+        the items to reduce
+
+    Returns
+    -------
+    dict
```
mibaumgartner: Applies to alle reduction functions: doen't seem correct when `items` is a list.

delira/logging/base_logger.py
```diff
@@ -2,6 +2,179 @@
...
+    Reduction Function returning the mean element
+
+    Parameters
+    ----------
+    items : list
+        the items to reduce
+
+    Returns
+    -------
+    dict
```
mibaumgartner: `np.mean` returns a ndarray.

delira/logging/base_logger.py
```diff
@@ -2,6 +2,179 @@
...
+    for k, v in d.items():
+        new_key = parent_key + sep + k if parent_key else k
+        if isinstance(v, MutableMapping):
+            items.extend(_flatten_dict(v, new_key, sep=sep).items())
+        else:
+            items.append((new_key, v))
+    return type(d)(items)
+
+
+def _unflatten_dict(dictionary, sep="."):
```
mibaumgartner: Is there a test for this function?

delira/logging/base_logger.py
```diff
@@ -2,6 +2,179 @@
...
+        d = return_dict
+        for part in parts[:-1]:
+            if part not in d:
+                d[part] = dict()
+            d = d[part]
+        d[parts[-1]] = value
+    return return_dict
+
+
+def _reduce_dict(items: list, reduce_fn):
```
mibaumgartner: Is there a test for this function?

delira/logging/base_logger.py
```diff
@@ -2,6 +2,179 @@
...
+        new_key = parent_key + sep + k if parent_key else k
+        if isinstance(v, MutableMapping):
+            items.extend(_flatten_dict(v, new_key, sep=sep).items())
+        else:
+            items.append((new_key, v))
+    return type(d)(items)
+
+
+def _unflatten_dict(dictionary, sep="."):
+    return_dict = {}
```
mibaumgartner: Docstring missing

delira/logging/base_logger.py
```diff
@@ -2,6 +2,179 @@
...
+    Returns
+    -------
+    dict
+        reduced items
+
+    """
+    return np.min(items)
+
+
+def _flatten_dict(d, parent_key='', sep='.'):
```
mibaumgartner: Is there a test for this function?

delira/logging/base_logger.py
```diff
@@ -2,6 +2,179 @@
...
+    items = []
```
mibaumgartner: Docstring missing.

delira/logging/base_logger.py
```diff
@@ -22,15 +196,101 @@ def __init__(self, backend: BaseBackend, max_queue_size: int = None,
...
+                or FunctionType. Specifies the logging value per key.
+                Missing keys will be filles with a default value of 'last'.
+                Valid types for strings are
+                'last' | 'first' | 'mean' | 'max' | 'min'.
         level : int
             the logging value to use if passing the logging message to
             python's logging module because it is not appropriate for logging
             with the assigned logging backend
+
+        Warnings
```
mibaumgartner, line 57: How do we want to add warnings to our docstrings? I started adding `.. warning :: ` sections to the description because those are later highlighted in the docs. (just checked the numpy docstring guide and they mention both options)

delira/logging/base_logger.py
```diff
@@ -22,15 +196,101 @@ def __init__(self, backend: BaseBackend, max_queue_size: int = None,
             the maximum size for the queue; if queue is full, all additional
             logging tasks will be dropped until some tasks inside the queue
             were executed; Per default no maximum size is applied
+        logging_frequencies : int or dict
```
mibaumgartner, line 29: Do we have a test for the logging frequency?

delira/models/abstract_network.py
```diff
@@ -51,8 +51,8 @@ def __call__(self, *args, **kwargs):
 
     @staticmethod
     @abc.abstractmethod
-    def closure(model, data_dict: dict, optimizers: dict, losses=None,
-                metrics=None, fold=0, **kwargs):
+    def closure(model, data_dict: dict, optimizers: dict, losses: dict,
+                fold=0, **kwargs):
```
mibaumgartner: The `BaseTrainer` uses `batch_nr`, do we want to add this as an explicit argument?

delira/training/base_experiment.py
```diff
@@ -122,9 +122,9 @@ def __init__(self,
         self.predictor_cls = predictor_cls
 
         if val_score_key is None:
-            if params.nested_get("val_metrics", False):
+            if params.nested_get("metrics", False):
```
mibaumgartner: I think the docstrings do not mention that the first metric will be used by default. I'm not even sure, if this is wanted 🙈

delira/training/base_experiment.py
```diff
@@ -205,7 +205,7 @@ def _setup_training(self, params, **kwargs):
         lr_scheduler_cls = training_params.nested_get("lr_sched_cls", None)
         lr_scheduler_params = training_params.nested_get("lr_sched_params",
                                                          {})
-        val_metrics = training_params.nested_get("val_metrics", {})
+        metrics = training_params.nested_get("metrics", {})
```
mibaumgartner, line 208: There should be an indication that the key was renamed to `metrics`.

delira/logging/base_logger.py
```diff
@@ -2,6 +2,179 @@
...
+    items : list
+        the items to reduce
+
+    Returns
+    -------
+    dict
+        reduced items
+
+    """
+    return items[-1]
```
mibaumgartner: Probably, should add a test to increase coverage....

delira/training/base_trainer.py
```diff
@@ -37,14 +32,14 @@ class BaseNetworkTrainer(Predictor):
                        "output_device",
                        "_callbacks"]
 
+    # TODO: Add new logging arguments to other trainers!
```
mibaumgartner: There is still a todo here

delira/training/base_trainer.py
```diff
@@ -59,6 +54,9 @@ def __init__(self,
                  metric_keys=None,
                  convert_batch_to_npy_fn=lambda x: x,
                  val_freq=1,
+                 logging_callback_cls=DefaultLoggingCallback,
```
mibaumgartner: Eventually reorder arguments, such that all logging related arguments are together.

delira/training/base_trainer.py
```diff
@@ -310,18 +370,31 @@ def _train_single_epoch(self, batchgen: Augmenter, epoch,
             iterable = enumerate(batchgen)
 
         for batch_nr, batch in iterable:
```
mibaumgartner: I would prefer consistent naming: either `batch_nr` or `iter_num` (so closure and callback have consistent names for the same thing)

delira/logging/base_logger.py
mibaumgartner: The reduce functions are the same as defined inside the `BaseTrainer`, I think we should have one place where we define them (and the dictionary which maps the strings to the functions) and import them from there.

delira/training/base_trainer.py
```diff
@@ -394,83 +467,63 @@ def reduce_fn(batch):
...
+            _, total_metrics = self._convert_to_npy_fn(**total_metrics)
+
+            for k, v in total_metrics.items():
+                total_metrics[k] = reduce_fn(v)
+
+            # check if metric became better
+            if val_score_key is not None:
+                if val_score_key not in total_metrics:
+                    if "val_" + val_score_key not in total_metrics:
+                        logger.warning(
```
mibaumgartner: Maybe change this to a normal warning (I think the warning should be printed to the terminal independent of the logger configuration)?

delira/training/base_trainer.py
```diff
@@ -136,8 +152,7 @@ def __init__(self,
         assert isinstance(save_path, str)
         assert isinstance(losses, dict)
         assert isinstance(optimizer_params, dict)
-        assert isinstance(train_metrics, dict)
-        assert isinstance(val_metrics, dict)
+        assert isinstance(metrics, dict)
```
mibaumgartner: Is there somewhere proper type-checking? (this will raise an `AssertionError` instead of a `ValueError`)

delira/training/base_trainer.py
```diff
@@ -685,14 +705,47 @@ def name(self):
...
+        ----------
+        logging_type : str or callable
+            the type of logging. If string: it must be one of
+            ["visdom", "tensorboardx"]
+            If callable: it must be a logging handler backend class
+        logging_kwargs : dict
+            dictionary containing all logging keyword arguments
+        logging_callback_cls : class
+            the callback class to create and register for logging
+        logging_callback_cls : class
```
mibaumgartner: duplicate `logging_callback_cls ` in docstring

delira/training/base_trainer.py
```diff
@@ -685,14 +705,47 @@ def name(self):
...
+            the callback class to create and register for logging
+        logging_frequencies : int or dict
+                specifies how often to log for each key.
+                If int: integer will be applied to all valid keys
+                if dict: should contain a frequency per valid key. Missing keys
+                will be filled with a frequency of 1 (log every time)
+                None is equal to empty dict here.
+        reduce_types : str of FunctionType or dict
```
mibaumgartner, line 771: Maybe be a little be more specific: what is reduced inside the logger (also applies to the docstring in the logger)?

delira/training/base_trainer.py
```diff
@@ -802,3 +853,14 @@ def _search_for_prev_state(path, extensions=None):
             return latest_state_path, latest_epoch
 
         return None, 1
+
+    def register_callback(self, callback: AbstractCallback):
```
mibaumgartner, line 901: Docstring missing

delira/training/base_trainer.py
```diff
@@ -802,3 +853,14 @@ def _search_for_prev_state(path, extensions=None):
...
+        has_all_attrs = True
+        for attr in ("iter", "epoch",):
```
mibaumgartner: `iter` is already checked in super class, do we want to check it here again.

delira/training/callbacks/abstract_callback.py
```diff
@@ -59,3 +59,41 @@ def at_epoch_end(self, trainer, **kwargs):
 
         """
         return {}
+
+    def at_iter_begin(self, trainer, **kwargs):
```
mibaumgartner: Explicitly define which arguments are passed here.

delira/training/callbacks/abstract_callback.py
```diff
@@ -59,3 +59,41 @@ def at_epoch_end(self, trainer, **kwargs):
...
+        Returns
+        -------
+        dict
+            modified trainer attributes, where the name must correspond to the
+            trainer's attribute name
+
+        """
+        return {}
+
+    def at_iter_end(self, trainer, **kwargs):
```
mibaumgartner: Explicitly define which arguments are passed here.

delira/training/callbacks/logging_callback.py
```diff
@@ -0,0 +1,85 @@
...
+import logging
+
+
+class DefaultLoggingCallback(AbstractCallback):
+    """
+    A default Logging backend which logs only the metrics; Should be
+    subclassed for additional logging
+    """
+
+    def __init__(self, backend: BaseBackend, max_queue_size: int = None,
```
mibaumgartner, line 12: Can we define `at_epoch_end` and log the reduced (typically average) values of the metrics? Or is this done somewhere else?

delira/training/predictor.py
```diff
@@ -547,8 +627,36 @@ def calc_metrics(batch: LookupConfig, metrics=None, metric_keys=None):
...
+        Parameters
+        ----------
+        callback : :class:`AbstractCallback`
+            the callback to register
+
+        Raises
+        ------
+        AssertionError
+            `callback` is not an instance of :class:`AbstractCallback` and has
+            not both methods ['at_epoch_begin', 'at_epoch_end']
```
mibaumgartner: Wrong fucntions in docstring: function checks if `at_iter_being/end` are present.

mibaumgartner (changes requested):
Added some comments and questions 🙈

tests/utils/test_messenger.py
mibaumgartner: The `DummyTrainer` is a little bit too dummy and had some problems with pickling. I overwrote the function because it doesn't need to save a real state (the state doesn't even change). The only reason to create the dummy is the internal state of the trainer for the callbacks.

delira/logging/base_logger.py
justusschock: This function was tested implicitly, because the logging would fail otherwise, so no, there isn't

delira/logging/base_logger.py
justusschock: Same as above

delira/training/base_trainer.py
justusschock: currently there isn't, but if you want to change this, feel free to do so.

Note: If you change this, I'd prefer to raise a `TypeError` and not a `ValueError`

delira/training/base_trainer.py
justusschock: No, we don't

delira/training/callbacks/abstract_callback.py
justusschock: This might be difficult, since this should work for all backends/arguments/callbacks. The only constant argument here is the trainer

delira/training/callbacks/abstract_callback.py
justusschock: Same as above

delira/training/callbacks/logging_callback.py
justusschock: This should not be necessary. It should be done on a per-iter basis combined with the corresponding frequency

delira/training/callbacks/abstract_callback.py
mibaumgartner: Okey, I understand your argument. Our `BaseTrainer` and `Predictor` currently pass some argument to the callbacks. These arguments should be mentioned somewhere, so you don't have to look inside the code of the trainer in order to determine which arguments are inside `kwargs`. I would propose to adjust the docstring with a section where the arguments are mentioned. 
https://github.com/delira-dev/delira/blob/d17809739945de7f977aef59c2947323716d467e/delira/training/predictor.py#L178
https://github.com/delira-dev/delira/blob/d17809739945de7f977aef59c2947323716d467e/delira/training/predictor.py#L203-L206

delira/training/callbacks/abstract_callback.py
mibaumgartner: I'm closing this because it is the same as `at_iter_begin` .

delira/training/callbacks/logging_callback.py
mibaumgartner: Currently the reduced values are logged and I liked the feature because it is very easy to see which epoch is the best.

delira/logging/base_logger.py
mibaumgartner: Okey, do we want explicit tests for this? ^^ Might be easier to identify the error in later stages ;)

delira/logging/base_logger.py
justusschock: Implicitly Tested -> No explicit test

delira/logging/base_logger.py
justusschock: A separate Warnings section (as done here) should also be highlighted and is more conform with the rest of the numpy doc format, which is why I prefer this way.

delira/logging/base_logger.py
justusschock: Not yet

delira/training/base_experiment.py
justusschock: This behavior was chosen to select a best model, if one forgot to provide the key (and in many cases there is only one metric calculated during training/validation). But we can also remove this

delira/training/base_trainer.py
justusschock: Just forgot to delete it

delira/training/base_trainer.py
justusschock: I would like this, but this would introduce some breaking changes, which is why I haven't done this. One the other hand: We have lot's of breaking changes, so we might consider doing this now and therefore avoid breaking changes in the future

delira/training/base_trainer.py
justusschock: I prefer `iter_num`

delira/training/callbacks/logging_callback.py
justusschock: So you want to log them in Addition to the ones with the frequency? We can certainly do this. What prefix/infix/suffix do you propose to distinguish them?

delira/training/base_experiment.py
mibaumgartner: I would like to remove it and insert a warning that no `val_score_key ` was set (so the user gets a notification and needs to take care for the correct behaviour).

delira/training/callbacks/logging_callback.py
mibaumgartner: Yes, I want them in addition to the ones with frequency 👍 I like `reduced_` as a prefix because the exact `reduce_fn` can vary.

justusschock: I would prefer `reduced_REDUCTIONTYPE_` and use the string passed to the keyword mapping as REDUCTIONTYPE, but in general this seems to be finde

justusschock: This is not as easy, as I thought, since we would have to bypass the logging frequency for these values, which is a thing, that is not implemented inside the logger, since it is undesired behavior. Otherwise, we would only log the reduced values every `n` epochs (if  `n` is the given logging frequency)

justusschock: So @mibaumgartner and I discussed elsewhere and decided to maybe include this later, but not in this PR

mibaumgartner (changes requested):
After merging https://github.com/delira-dev/delira/pull/201 we could easily add a test to check for execution of the iter callback. The rest looks good to go.

mibaumgartner (approved):

PR #173, opened 2019-07-31
Documentation and Examples

justusschock:
This PR adds documentation and examples for the new backends (and updates old examples). 

This PR should be merged after #102 and #172 were merged


docs/index.rst
```diff
@@ -3,9 +3,9 @@
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
-===================================
-delira - Deep Learning In RAdiology
-===================================
+=====================================================================
+delira - delira - A Backend Agnostic High Level Deep Learning Library
```
mibaumgartner: Only one delira :)

notebooks/classification_examples/pytorch.ipynb
```diff
@@ -200,9 +163,131 @@
...
+    "class Flatten(torch.nn.Module):\n",
+    "        \n",
+    "    def forward(self, x):\n",
+    "        return x.view(x.size(0), -1)\n",
+    "\n",
+    "class SmallVGGPyTorch(AbstractPyTorchNetwork):\n",
+    "    def __init__(self, in_channels, num_classes):\n",
+    "        super().__init__()\n",
+    "        \n",
+    "        self.model = torch.nn.Sequential(\n",
```
mibaumgartner: Maybe we should use a simpler model so it is a bit easier for beginners?

mibaumgartner (approved):
Only minor remarks, rest looks good :D

notebooks/classification_examples/pytorch.ipynb
justusschock: That's why I opted for VGG and not for any Resnet like model. What do you suggest?

mibaumgartner: To demonstrate the general capability of training a NN with Delira we would only need a very small net. Some conv+pool and fc (around 3-4 layers) should be sufficient. Something like this https://www.tensorflow.org/tutorials/keras/basic_classification :)

mibaumgartner:
Furthermore, we should also add the prediction of a datamanger with `experiment.test()` to the notebooks. Maybe even something with `experiment.resume()` .

notebooks/classification_examples/pytorch.ipynb
justusschock: Okay, I'll change this

justusschock:
I don't know if this wouldn't be too much for the notebooks, we should maybe leave these notebooks as they are (after simplifying the model structure) and create another one containing a whole pipeline (with only one backend though).

mibaumgartner:
Ok, we can do that, too :) 
Will you add the new notebook to this PR?

justusschock:
Probably not, I will only simply the models in the existing notebooks and then this should be good to go. Otherwise this PR becomes way too huge

justusschock:
@mibaumgartner is the architecture now simple enough?

notebooks/classification_examples/chainer.ipynb
```diff
@@ -166,8 +166,8 @@
     "from batchgenerators.transforms.sample_normalization_transforms import MeanStdNormalizationTransform\n",
     "\n",
     "transforms = Compose([\n",
-    "    RandomCropTransform(200), # Perform Random Crops of Size 200 x 200 pixels\n",
-    "    ResizeTransform(224), # Resample these crops back to 224 x 224 pixels\n",
+    "    RandomCropTransform(24), # Perform Random Crops of Size 200 x 200 pixels\n",
```
mibaumgartner: Please adjust the comments accordingly.

notebooks/classification_examples/pytorch.ipynb
```diff
@@ -118,8 +118,8 @@
     "from batchgenerators.transforms.sample_normalization_transforms import MeanStdNormalizationTransform\n",
     "\n",
     "transforms = Compose([\n",
-    "    RandomCropTransform(200), # Perform Random Crops of Size 200 x 200 pixels\n",
-    "    ResizeTransform(224), # Resample these crops back to 224 x 224 pixels\n",
+    "    RandomCropTransform(24), # Perform Random Crops of Size 200 x 200 pixels\n",
+    "    ResizeTransform(32), # Resample these crops back to 224 x 224 pixels\n",
```
mibaumgartner: Please adjust the comments accordingly.

justusschock:
So, after I did this, it is ready to merge? Or do you want additional simplifications?

mibaumgartner:
Should be ready to merge, rest looks good and simple :D

PR #182, opened 2019-08-10
Fix notebooks and trainer indent

mibaumgartner:
- [x] Trainer indentation bug

Notebooks:
- [x] Torch
- [ ] Torchscript [can not save model]
- [ ] Tf Eager [something with the dense layer in the model]
- [ ] Tf Graph [something model related, may have wrong return args in closure]
- [ ] Chainer [something related to input format; may have wrong return args in closure]
- [ ] Sklearn [too many dimensions in last cell (manual prediction)]

### DIFF
--- a/delira/training/base_trainer.py
+++ b/delira/training/base_trainer.py
@@ -431,7 +431,7 @@ class BaseNetworkTrainer(Predictor):
                 total_metrics = {
                     **train_metrics,
                     **train_losses}
-                
+
                 # validate network
                 if datamgr_valid is not None and (epoch % self.val_freq == 0):
                     # next must be called here because self.predict_data_mgr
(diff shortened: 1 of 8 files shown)


ORippler:
While fixing tf_eager notebook, I noticed 2 things:

1. tf uses one_hot encoding for it's losses, and i think torch uses labelencoding. Some kind of make_one_hot function would be nice (either as a batchgenerators transform or inside the ClassificationFakeData class.) That's why the achieved/reported accuracy is meaningless, as I hacked a working solution.

2. Batchgenerators transforms all only support channels_first format. So we should issue some kind of warning for tf or do automated transposition if channels_last is specified.

mibaumgartner:
1. Delira Vision contains one hot encodings for numpy https://github.com/delira-dev/vision_base/blob/f4952e80a41824ed1ee713d587ad92b915f787cc/deliravision/utils/tensor_ops.py#L4-L6 and pytorch https://github.com/delira-dev/vision_torch/blob/78f8fac388e9fcfd369b5976df8433e5924cac98/deliravision/torch/utils/tensor_ops.py#L4 .
I think, its a good idea to create additional transformations because their are so commonly used. 

2. I like the idea :) Is there a possibility to detect if the tensorflow backend is used for training? Should we create an additional transformation for covnerting between channel first and last (and reverse)?

justusschock:
The second part could most easily be done inside the network, since there is also a data format present and the permutation itself is not a time consuming operation rather than changing some internal references. 

I agree, that maybe the numpy part should be part of the main repo (and wrapped as a transform). We would only need the numpy part here, since we should convert it, before moving to the native tensor type.

PR #184, opened 2019-08-21
Argument Resolution Refactoring

### LINKED ISSUE #180: Argument Resolution Refactoring
We should definately check where to give what argument (and maybe expose more keyword arguments) in experiment (Possible ways are to pass arguments via keyword or parameter object in `__init__.py` or `run`/`test`/`resume`/`kfold`, so there are 4 different ways to pass arguments)

We should probably do this, after merging #166

justusschock:
Fixes #180 once it's done.

Tests have not yet been updated. Before doing this, a feedback on the general structure would be nice


delira/training/backends/torch/experiment.py
```diff
@@ -8,103 +7,240 @@
...
+            of 'highest' | 'lowest'
+        val_freq : int
+            specifies how often to run a validation step
+        callbacks : list
+            list of callbacks to use during training. Each callback should be
+            derived from :class:`delira.training.callbacks.AbstractCallback`
+        trainer_cls : type
+            the class implementing the actual training routine
         **kwargs :
             additional keyword arguments
```
mibaumgartner: Please mention where the kwargs are going.

delira/training/backends/torch/experiment.py
```diff
@@ -8,103 +7,240 @@
...
+            val_score_key=val_score_key, val_score_mode=val_score_mode,
+            val_freq=val_freq, callbacks=callbacks,
+            trainer_cls=trainer_cls, **kwargs)
+
+    def test(self, model, test_data, prepare_batch, callbacks=None,
+             predictor_cls=Predictor, metrics=None, metric_keys=None,
+             **kwargs):
+        """
+        Setup and run testing on a given network
+        Parameters
```
mibaumgartner, line 142: Blank line between description and parameters missing.

delira/training/backends/torch/experiment.py
```diff
@@ -8,103 +7,240 @@
...
+        ----------
+        model : :class:`AbstractNetwork`
```
mibaumgartner, line 144: It would be great if the order of the argument descriptions is the same as the order in the call (make life a little bit easier ^^ )

delira/training/backends/torch/experiment.py
```diff
@@ -8,103 +7,240 @@
...
+        prepare_batch : function
+            function to convert a batch-dict to a format accepted by the model.
+            This conversion typically includes dtype-conversion, reshaping,
+            wrapping to backend-specific tensors and pushing to correct devices
+        callbacks : list
+            list of callbacks to use during training. Each callback should be
+            derived from :class:`delira.training.callbacks.AbstractCallback`
+        predictor_cls : type
+            the class implementing the actual prediction routine
+        **kwargs :
```
mibaumgartner, line 166: Please mention where the kwargs are going.

delira/training/backends/torch/experiment.py
```diff
@@ -8,103 +7,240 @@
...
+        test_data : :class:`BaseDataManager`
+            the data to use for testing
+        metrics : dict
+            the metrics to calculate
+        metric_keys : dict of tuples
+            the batch_dict keys to use for each metric to calculate.
+            Should contain a value for each key in ``metrics``.
+            If no values are given for a key, per default ``pred`` and
+            ``label`` will be used for metric calculation
+        prepare_batch : function
```
mibaumgartner, line 148: Comment that the default will be set to the model's `prepare_batch` function

delira/training/backends/torch/experiment.py
```diff
@@ -114,29 +250,23 @@ def kfold(self, data: BaseDataManager, metrics: dict, num_epochs=None,
...
-            the training and model parameters
-            (will be merged with ``self.params``)
-        verbose : bool
-            verbosity
+        prepare_batch : function
+            function to convert a batch-dict to a format accepted by the model.
+            This conversion typically includes dtype-conversion, reshaping,
+            wrapping to backend-specific tensors and pushing to correct devices
         **kwargs :
             additional keyword arguments
```
mibaumgartner: There are no kwargs in the function call.

mibaumgartner (changes requested):
I reviewed the pytorch experiment and the baseexperiment (most of the comments should apply to the other experiments, too). Really digging the new argument structure, much more intuitive. Added a bit more comment than usual because this is a good opportunity to improve the general interface of delira.

Further considerations include:
1. Where is the batchsize determined. After some thought, I really liked the idea of you to use the batchsize from the datamanager.

2. Where is a good point to mention which arguments need to be passed by the parameters object | datamanager | experiment ? (the experiment args are obvious but the parameters and datamangers aren't)

3. The notebooks should be updated accordingly, too. (This should probably happen after https://github.com/delira-dev/delira/pull/182) 

4. Do we want to include typings for the experiment interfaces? (I think this would be the right to do that)

justusschock:
All the changes you requested are making sense to me, so I'll add them later on.

Regarding your considerations:

1. The batchsize is only determined during initialization of the datamanager, so this does not involve this part (Of course one could use the parameters class to temporarily store this value too).

2. Probably inside our basic `delira` tutorial, as well as in the desciption of our `BaseExperiment`?

3. We should definitely do this. I would prefer to integrate it into #182 (after this PR is merged)instead of adding it as a separate PR.

4. I would not add the typing here, but do it, while we add stub files (and check the whole project's typing)

PR #185, opened 2019-08-25
Explicitly free memory of old preds_batch

mibaumgartner:
Fixes a memory leak during the prediction phase during training.


justusschock (approved):
Seems good to go

PR #192, opened 2019-08-30
fix missing voxel geometry info after resampling

cwern1:


justusschock:
Can you maybe add this to the [vision-base](https://github.com/delira-dev/vision_base) repo? Since these are vision based things, we are currently working on removing them from this repo.

PR #193, opened 2019-08-30
Add compatibility to newer versions of slackclient

mibaumgartner:
Add compatibility to newer versions of slackclient >= 1.3.1


justusschock (approved):

justusschock:
The only thing is to remove the 1.3.1 from the dependency in the docstring

PR #201, opened 2019-09-07
Add training callbacks

mibaumgartner:
- [x] Add training callbacks 
- [x] Adjust doc string  of callback
- [x] Add test for callback


justusschock (approved):
Looks good to go, once the tests passed

PR #206, opened 2019-09-09
Remove trixi

mibaumgartner:
Introduces a new config class to save all delira related parameters. 
Furthermore, `trixi` and `trixi-slim` are removed from the dependencies.


delira/utils/codecs.py
```diff
@@ -0,0 +1,539 @@
...
+import numpy as np
+import logging
+import typing
+from functools import partial
+import typing
+
+
+class Encoder:
+    """
+    Encode arbitrary objects as dicts, str, int, float, list
```
justusschock: What about tuples and arrays?

delira/utils/__init__.py
```diff
@@ -1,10 +1,6 @@
...
-from delira.utils.imageops import bounding_box
-from delira.utils.imageops import calculate_origin_offset
-from delira.utils.imageops import max_energy_slice
-from delira.utils.imageops import sitk_new_blank_image
-from delira.utils.imageops import sitk_resample_to_image
-from delira.utils.imageops import sitk_resample_to_shape
-from delira.utils.imageops import sitk_resample_to_spacing
-from delira.utils.path import subdirs
-from delira.utils.time import now
+from .config import DeliraConfig, Config
```
justusschock: No relative import paths

delira/utils/codecs.py
justusschock: What about other types as tuples and arrays?

delira/utils/codecs.py
```diff
@@ -0,0 +1,539 @@
...
+        Encode arbitrary objects as dicts, str, int, float, list
+
+        Parameters
+        ----------
+        obj : Any
+            object to be encoded
+
+        Returns
+        -------
+        typing.Any
```
justusschock: omit the `typing` part in returntype docstrings (applies for all docstrings, I only add this as a comment here), since this is not a real type rather than a placeholder

delira/utils/codecs.py
```diff
@@ -0,0 +1,539 @@
...
+            encoded object
+        """
+        # use type() to check for dict and list because type() does not
+        # consider subtypes which is the desired behaviour in this case
+        if isinstance(obj, (str, int, float)):
```
justusschock, line 50: you have to redo this (probably due to autopep8); We might have to look for a more beautiful way to check this (compliant with pep8) or turn off this type of conversions

delira/utils/codecs.py
```diff
@@ -0,0 +1,539 @@
...
+        obj : list
+            list to be encoded
+
+        Returns
+        -------
+        list
+            list with encoded internal items
+        """
+        # if list: add explicit list specifier
+        return [self.encode(i) for i in obj]
```
justusschock, line 93: there is no explicit list specifier here? Either change this or remove/change comment

delira/utils/codecs.py
```diff
@@ -0,0 +1,539 @@
...
+        ----------
+        obj : dict
+            dict to be encoded
+
+        Returns
+        -------
+        dict
+            dict with encoded internal items
+        """
+        # if dict: add specific dict specifier
```
justusschock: there is no explicit dict specifier here? Either change this or remove/change comment

delira/utils/codecs.py
```diff
@@ -0,0 +1,539 @@
...
+        if not isinstance(classargs, dict):
+            raise TypeError("Arguments for classargs must be defined as dict.")
+
+        obj_cls = getattr(importlib.import_module(classargs["module"]),
+                          classargs["name"])
+        args = classargs.get("args", [])
+        kwargs = classargs.get("kwargs", {})
+        return obj_cls(*args, **kwargs)
+
+    def _decode_functionargs(self, obj: dict) -> typing.Any:
```
justusschock, line 508: does this also work with classes to create a class instance?

delira/utils/config.py
```diff
@@ -1,13 +1,509 @@
...
+        ----------
+        dict_like : dict, optional
+            dict like object to initialize config, by default None
+        kwargs:
+            additional arguments added to the config
+
+        Warnings
+        --------
+        If the short form of the nested settings is used integers and str
+        can not be differentiated. Therefore, integers should be avoided.
```
justusschock: don't specify this to ints, but more general to other (non-string) types as floats etc. are a problem too

delira/utils/config.py
```diff
@@ -1,13 +1,509 @@
...
+        key : str
+            key for desired value
+
+        Returns
+        -------
+        bool
+            true if key is in config
+        """
+        contain = True
+        try:
```
justusschock, line 237: Do we only want to resolve the key like this or with `nested_get` to also check it's presence in subdicts?

delira/utils/config.py
```diff
@@ -1,13 +1,509 @@
...
+        """
+        current_level = self
+        for k in keys:
+            if k not in current_level:
+                current_level[k] = self._create_internal_dict()
+            # traverse to needed dict
+            current_level = current_level._try_key(k)
+        return current_level
+
+    def _try_key(self, key):
```
justusschock: I don't know, if we really should do it like this, I'd prefer to use strings and only raise warnings otherwise

delira/utils/config.py
```diff
@@ -1,13 +1,509 @@
...
+
+        Raises
+        ------
+        ValueError
+            if overwrite is not enabled and `update_dict` contains same values
+            as config
+        """
+        for key, item in update_dict.items():
+            # check for overwrite
+            self._raise_overwrite(key, overwrite=overwrite)
```
justusschock: I think it would be cleaner to do this in `self._update` since we then had a function dealing with the dict as whole and mapping only a single function to all its items

delira/utils/config.py
```diff
@@ -1,13 +1,509 @@
...
+    def _raise_overwrite(self, key, overwrite):
+        """
+        Checks if a ValueError should be raised
+
+        Parameters
+        ----------
+        key : str
+            key which needs to be checked
+        overwrite : bool
+            if overwrite ir enables no error is raised even if key already
```
justusschock: typos

delira/utils/config.py
```diff
@@ -1,13 +1,509 @@
...
+            path where config is saved
+        formatter : callable, optional
+            defines the format how the config is saved, by default yaml.dump
+        encoder_cls : :class:`Encoder`, optional
+            trasforms config to a format which can be formatted by the
+            `formatter`, by default Encoder
+        kwargs:
+            additional keyword arguments passed to `formatter`
+        """
+        self._timestamp = now()
```
justusschock, line 342: Maybe we should initialize this with None and only change the timestamp if it hasn't been set to anything different so far?

delira/utils/config.py
```diff
@@ -1,13 +1,509 @@
...
+        ----------
+        path : str
+            path where config is saved
+        formatter : callable, optional
+            defines the format how the config is saved, by default yaml.dump
+        encoder_cls : :class:`Encoder`, optional
+            trasforms config to a format which can be formatted by the
+            `formatter`, by default Encoder
+        kwargs:
+            additional keyword arguments passed to `formatter`
```
justusschock: use ```:param`formatter` ``` instead (Applies to all docstrings, only mentioned here)

delira/utils/config.py
```diff
@@ -1,13 +1,509 @@
...
+        """
+        self._timestamp = now()
+        encoded_self = encoder_cls().encode(self)
+        return formatter(encoded_self, **kwargs)
+
+    @classmethod
+    def load(cls, path, formatter=yaml.load, decoder_cls=Decoder, **kwargs):
```
justusschock: I am not sure, if I want this to be a class method.

What about making this a usual method and calling self.update on the loaded part and making a separate class method (something like `from_file`), which creates the class and calls this method? I think this would be a bit cleaner?

delira/utils/config.py
```diff
@@ -1,13 +1,509 @@
...
+            additional keyword arguments passed to `formatter`
+        """
+        decoded_format = formatter(data, **kwargs)
+        decoded_format = decoder_cls().decode(decoded_format)
+        if not isinstance(decoded_format, cls):
+            decoded_format = cls(decoded_format)
+        return decoded_format
+
+    @classmethod
+    def create_from_dict(cls, value, deepcopy=False):
```
justusschock, line 408: What's the difference from calling `Config.create_from_dict(a_dict)` to simply `Config(a_dict)`? If there is a difference, this should be stated somethere, else this function is not necessary

delira/utils/config.py
```diff
@@ -1,13 +1,509 @@
...
+
+        Raises
+        ------
+        TypeError
+            if value is not an instance of :class:`ArgumentParser`
+            or :class:`Namespace`
+        """
+        if isinstance(value, argparse.ArgumentParser):
+            args_parsed = value.parse_args(**kwargs)
+            return cls.create_from_dict(vars(args_parsed), deepcopy=deepcopy)
```
justusschock: it would be cleaner to call `cls.create_from_argparse(args_parsed. deepcopy=deepcopy)` . If we need to change the `vars` path, we only have to do it in one line, making it easier to maintain. And I think the overhead of an additional function call is negligible

delira/utils/config.py
```diff
@@ -51,20 +547,219 @@ def nested_get(self, key, *args, **kwargs):
...
+
+        Returns
+        -------
+        Any
+            dict if value is none otherwise value is returned
+        """
+        if value is None:
+            return {}
+        else:
+            return value
```
justusschock: maybe add an explicit conversion to `dict` here? This would make possible errors easier to track down since it's failure would be more clear.

delira/utils/config.py
```diff
@@ -51,20 +547,219 @@ def nested_get(self, key, *args, **kwargs):
...
+        """
+        # create empty dict
+        if "model" not in new_params:
+            new_params["model"] = {}
+
+        # create empty dict
+        if "training" not in new_params:
+            new_params["training"] = {}
+
+        self.variable_model.update(new_params["model"], overwrite=True)
```
justusschock: Should this also be part of the general Config? I thought a clean way to do this, would be to maintain 4 lists of keys and then build a dict of the config entries in the property method? This would keep the objects in the config without the need to copy them, but would require some additional book-keeping inside the `pop` and `setitem` methods

justusschock (changes requested):
I left some comments, nothing major, but some points to discuss and some slight changes

delira/utils/codecs.py
mibaumgartner: The docstring is a bit misleading and I will change it. The correct meaning would be like:
"The encoded object only consists of dicts, str, int, float, list" .

delira/utils/codecs.py
mibaumgartner: I did not find a different solution for this problem (before the autopep8 fix). The `isinstance() and not isinstance()` construct only works for specified subclasses (these are named in the nor instance part) and is not generally applicable to subclasses.

delira/utils/codecs.py
mibaumgartner: Gonna remove the docstring. This was from an old version.

delira/utils/codecs.py
mibaumgartner: same as above

delira/utils/codecs.py
mibaumgartner: There is a second function called `_decode_classargs` which can be used to create class instances. `_decode_functionargs` is only intended to create functions and set arguments with `partial`.

delira/utils/config.py
mibaumgartner: We could add this to the `LookupConfig`. The default config does not support `nested_get`.

delira/utils/config.py
mibaumgartner: We can do this. I removed it from the `__init__` because it would overwrite the original timestamp if the config was loaded.

delira/utils/config.py
mibaumgartner: This would imply an extra instruction each time a config needs to be loaded (because the config must be instantiated first and can only be loaded in a second step). 
Do we want this?

delira/utils/config.py
mibaumgartner: The is no difference. I preferred the syntax (it is also consistent with the `create_from_argparse` ) but ran into problems, because dicts need to support conversion of input parameters inside the `__init__` function. That's why there are two ways to do the same thing. 
If you prefer removing it, I can do that.

delira/utils/config.py
mibaumgartner: I don't completely understand what you mean. argparse.ArgumentParser.parse_args() returns an argparse.Namespace object which needs to be used to create the config. Is there an other way to achieve this without vars ?

delira/utils/config.py
justusschock: But in the `__init__` you would first initialize it with `None` and then overwrite it with the loaded values

delira/utils/config.py
justusschock: As long as this isn't a user instruction, this should be fine I guess. This way, we also have the option to update from file

delira/utils/config.py
justusschock: I'm fine with keeping it, but shouldn't we simply call the init here after the assert with the value dict?

delira/utils/config.py
justusschock: No, there isn't, but you could rewrite this to 
```
if isinstance(value, argparse.ArgumentParser):
            args_parsed = value.parse_args(**kwargs)
            return cls.create_from_argparse(args_parsed)
elif isinstance(value, argparse.Namespace):
            return cls.create_from_dict(vars(value), deepcopy=deepcopy)
else:
            raise TypeError("Type of args not supported.")
```
which would be a bit cleaner if we might want to refactor the Namespace part sometime in the future

delira/utils/config.py
mibaumgartner: I think we could just remove the setter function completely and use the getter to construct the necessary structure of the parameters?

delira/utils/config.py
mibaumgartner: You are right, I will change that.

delira/utils/config.py
justusschock: I'm afraid that we can't since without a setter, we don't know what arguments are part of the params and to which param-type they belong

delira/utils/config.py
mibaumgartner: I tried adding it to the `__init__` and the result was too messy because every internal dictionary is a config itself and thus has its own timestamp. Do we want a separate argument to add the timestamp to the config at initialisation or do we leave it like the current implementation (added during dumping the config)?

delira/utils/codecs.py
mibaumgartner: I tried `# nopep8` but it did not work for me so I disblaed E721 in `setup.cfg`

delira/utils/config.py
justusschock: No, we leave it as is. It was just a suggestion.

justusschock (approved):
This looks good to go. No changes necessary

Issue #214, opened 2019-09-15
[FeatureRequest] Logger registration


gedoensmax:
I think it would be nice to register the logger created by the experiment not only as callback but also as logger as it is not possible to log anything from outside the experiment which might be desired.

PR #215, opened 2019-09-16
Register logger

gedoensmax:
Concerns issue #214.
Could anyone have a look at the unittests i made ? 
I am not sure how to check if logging worked i tried to do the same as in logging frequency but that did not work.


mibaumgartner:
What is the exact problem you are facing when using the same approach as in the test of the logging freq?
I think you want to instantiate `DummyNetwork` and not `AbstractNetwork` for the trainer (maybe this solves your problem?)

gedoensmax:
This pull request also fixed a bug  where logging kwargs were not set to defaults.

mibaumgartner:
Forgot to add `train` argument to docstrings of iter callback.

mibaumgartner (approved):

PR #224, opened 2019-10-09
Fix lookup of non existent keys

mibaumgartner:
* Fix lookup of non existent keys 
* add option to allow nested lookup with multiple results (e.g. for `__contains__`)

### DIFF
--- a/delira/utils/config.py
+++ b/delira/utils/config.py
@@ -134,7 +134,11 @@ class Config(dict):
         current_level = self
         for k in keys:
             if k not in current_level:
-                current_level[k] = self._create_internal_dict()
+                if create:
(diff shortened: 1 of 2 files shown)


delira/utils/config.py
```diff
@@ -561,13 +565,15 @@ def nested_get(self, key, *args, **kwargs):
...
+        allow_multiple: bool
+            allow multiple results
         **kwargs :
             keyword arguments to provide default value
 
         Raises
         ------
         KeyError
-            Multiple Values are found for key
+            Multiple Values are found for key and allow_multiple is False
```
justusschock: NIT: Use :param:`allow_multiple` here

justusschock (approved):

PR #225, opened 2019-10-09
remove globalstep from logging

mibaumgartner:
Global step is also increased during validation which leads to funny looking skips in logging.
image


justusschock:
I'm not sure if we can simply remove this. I think some logging actions actually require the global step.

mibaumgartner:
I think, the current implementation of the default logging callback does not need the global step because there is only one place where logging takes place and each metric is only logged once per epoch. (when adding further values in irregular places the global step would be needed)
Edit: Alternatively, we could also use different counters for training and validation.

justusschock (approved):

PR #231, opened 2019-10-20
Update Config from system args

gedoensmax:
Make it possible to available an update for all arguments with one line of code. 
Although it is now possible to create an argparser based on a config.

### DIFF
--- a/delira/utils/config.py
+++ b/delira/utils/config.py
@@ -625,8 +625,9 @@ class Config(dict):
             params = vars(params)
             if len(unknown) > 0 and add_unknown_items:
                 warnings.warn(
-                    "Called with unknown arguments: {} They will note be stored "
-                    "if you do not set `add_unknown_items` to true.".format(unknown),
+                    "Called with unknown arguments: {} "
+                    "They will note be stored if you do not set "
+                    "`add_unknown_items` to true.".format(unknown),
                     RuntimeWarning)
 
             new_params = Config()
--- a/delira/utils/config.py
+++ b/delira/utils/config.py
@@ -32,7 +32,7 @@ def non_string_warning(func):
         if not isinstance(key, str):
             warnings.warn("The key {} is not a string, but a {}. "
                           "This may lead to unwanted behavior!".format(
-                key, type(key)), RuntimeWarning)
+                              key, type(key)), RuntimeWarning)
 
         return func(config, key, *args, **kwargs)
 
--- a/delira/utils/config.py
+++ b/delira/utils/config.py
@@ -30,7 +30,7 @@ def non_string_warning(func):
         if not isinstance(key, str):
             warnings.warn("The key {} is not a string, but a {}. "
                           "This may lead to unwanted behavior!".format(
-                key, type(key)), RuntimeWarning)
+                              key, type(key)), RuntimeWarning)
 
         return func(config, key, *args, **kwargs)
 
--- a/delira/utils/config.py
+++ b/delira/utils/config.py
@@ -7,6 +7,7 @@ from .codecs import Encoder, Decoder
 
 import yaml
 import argparse
+import sys
 
 
 def non_string_warning(func):
@@ -29,9 +30,10 @@ def non_string_warning(func):
         if not isinstance(key, str):
             warnings.warn("The key {} is not a string, but a {}. "
                           "This may lead to unwanted behavior!".format(
-                              key, type(key)), RuntimeWarning)
+                key, type(key)), RuntimeWarning)
 
         return func(config, key, *args, **kwargs)
+
     return warning_wrapper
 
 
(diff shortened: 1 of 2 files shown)


delira/utils/config.py
```diff
@@ -513,6 +515,135 @@ def create_from_str(cls, data, formatter=yaml.load, decoder_cls=Decoder,
...
+            else:
+                raise argparse.ArgumentTypeError('Boolean value expected.')
+
+        def add_val(dict_like, prefix=''):
+            for key, val in dict_like.items():
+                name = "--{}".format(prefix + key)
+                if val is None:
+                    parser.add_argument(name)
+                else:
+                    if type(val) == bool:
```
justusschock: can we change this to an instance check? Hard-Checks of types should be avoided if possible...
Also passing a function here as type is not a good idea for several reasons (as this usually should be a class and might not work with related instance-checks)

delira/utils/config.py
```diff
@@ -513,6 +515,135 @@ def create_from_str(cls, data, formatter=yaml.load, decoder_cls=Decoder,
...
+                        parser.add_argument(name, type=str2bool)
+                    elif isinstance(val, (list, tuple)):
+                        if len(val) > 0 and type(val[0]) != type:
+                            parser.add_argument(name, type=type(val[0]))
+                        else:
+                            parser.add_argument(name)
+                    elif isinstance(val, dict):
```
justusschock: Can you check for the more general `collections.Mapping?` here instead of `dict`?

delira/utils/config.py
```diff
@@ -513,6 +515,135 @@ def create_from_str(cls, data, formatter=yaml.load, decoder_cls=Decoder,
...
+
+        def add_val(dict_like, prefix=''):
+            for key, val in dict_like.items():
+                name = "--{}".format(prefix + key)
+                if val is None:
+                    parser.add_argument(name)
+                else:
+                    if type(val) == bool:
+                        parser.add_argument(name, type=str2bool)
+                    elif isinstance(val, (list, tuple)):
```
justusschock: if you move the check for dicts/Mappings before this one, you could check for the more general `collections.Iterable`

delira/utils/config.py
```diff
@@ -513,6 +515,135 @@ def create_from_str(cls, data, formatter=yaml.load, decoder_cls=Decoder,
...
+                        if len(val) > 0 and type(val[0]) != type:
+                            parser.add_argument(name, type=type(val[0]))
+                        else:
+                            parser.add_argument(name)
+                    elif isinstance(val, dict):
+                        add_val(val, prefix=key + '.')
+                    elif isinstance(val, type):
+                        parser.add_argument(name, type=val)
```
justusschock, line 548: I doubt that this would work, because you want to get a class-object here, but tell the argparser to expect an instance of the same class...

delira/utils/config.py
```diff
@@ -513,6 +515,135 @@ def create_from_str(cls, data, formatter=yaml.load, decoder_cls=Decoder,
...
+                    if type(val) == bool:
+                        parser.add_argument(name, type=str2bool)
+                    elif isinstance(val, (list, tuple)):
+                        if len(val) > 0 and type(val[0]) != type:
+                            parser.add_argument(name, type=type(val[0]))
+                        else:
+                            parser.add_argument(name)
+                    elif isinstance(val, dict):
+                        add_val(val, prefix=key + '.')
+                    elif isinstance(val, type):
```
justusschock: can you change this one to `elif isinstance(val, type) of inspect.isclass(val)`? There are some few cases, where classes don't inherit from `type`

delira/utils/config.py
```diff
@@ -513,6 +515,135 @@ def create_from_str(cls, data, formatter=yaml.load, decoder_cls=Decoder,
...
+
+        def str2bool(v):
+            if v.lower() in ('yes', 'true', 't', 'y', '1'):
+                return True
+            elif v.lower() in ('no', 'false', 'f', 'n', '0'):
+                return False
+            else:
+                raise argparse.ArgumentTypeError('Boolean value expected.')
+
+        def add_val(dict_like, prefix=''):
```
justusschock, line 532: a general question here? Shouldn't we use the actual `val` as a default argument here (it the argument type is not a dict)? this would allow us to optionally overwrite these args but to also retain the same config if flags are not specified

delira/utils/config.py
```diff
@@ -513,6 +515,135 @@ def create_from_str(cls, data, formatter=yaml.load, decoder_cls=Decoder,
...
+        Following the pattern: `--training.learning_rate 1234`
+
+        Returns
+        -------
+        argparse.ArgumentParser
+            parser for all variables in the config
+        '''
+        parser = argparse.ArgumentParser(allow_abbrev=False)
+
+        def str2bool(v):
```
justusschock: personally i don't like this approach. there are explicit actions inside argparsers for this: `store_true` and `store_false`. I also think, that we should only check for real boolean values here and not for the strings.

delira/utils/config.py
```diff
@@ -513,6 +515,135 @@ def create_from_str(cls, data, formatter=yaml.load, decoder_cls=Decoder,
...
+        ------
+        unknown_args : list
+            list of unknown args
+        Returns
+        ------
+        Config
+            a config of the parsed args
+        '''
+        # first element in the list must be a key
+        assert unknown_args[0].startswith('--')
```
justusschock: This should not raise an `AssertionError` but either a `ValueError` or a `RuntimeError`. Also you should first check whether `unknown_args[0]` is actually a string and else you would have to convert it first to make sure, it supports a `.startswith` method.

delira/utils/config.py
```diff
@@ -513,6 +515,135 @@ def create_from_str(cls, data, formatter=yaml.load, decoder_cls=Decoder,
...
+                    elif isinstance(val, type):
+                        parser.add_argument(name, type=val)
+                    else:
+                        parser.add_argument(name, type=type(val))
+
+        add_val(self)
+        return parser
+
+    @staticmethod
+    def _add_unknown_args(unknown_args):
```
justusschock, line 556: Can we somehow also enable support for flags like `-r` starting with only a single `-` as abbreviation for a `--`-flag?

delira/utils/config.py
```diff
@@ -513,6 +515,135 @@ def create_from_str(cls, data, formatter=yaml.load, decoder_cls=Decoder,
...
+            dictionary containing only updated arguments
+        '''
+
+        if len(sys.argv) > 1:
+            if not parser:
+                parser = self.create_argparser()
+
+            params, unknown = parser.parse_known_args()
+            params = vars(params)
+            if len(unknown) > 0 and add_unknown_items:
```
justusschock: I think this condition must be changed to `if unknown and not add_unknown_items` to reflect the actual state reflected in the warning below. (Note that there is now need to include `>0` or even the length method in this condition since `bool(len(iterable)>0) == bool(len(iterable)) == bool(iterable)`

tests/utils/test_config.py
```diff
@@ -215,6 +217,44 @@ def test_internal_type(self):
...
+        parser = cf.create_argparser()
+        known, unknown = parser.parse_known_args(testargs)
+        self.assertEqual(vars(known)['shallowNum'], 10)
+        self.assertEqual(vars(known)['deep.deepStr'], 'check')
+        self.assertEqual(unknown, ['--testlist', 'ele1', 'ele2', '--setflag'])
+
+    @unittest.skipUnless(
+        check_for_no_backend(),
+        "Test should only be executed if no backend is specified")
+    def test_update_from_argparse(self):
```
justusschock, line 244: Can you also check whether the warning is raised?

justusschock (changes requested):
Added some change-requests. Sometimes they are essential AFAIK, sometimes they could work but are likely to cause future problems and sometimes they are to improve the code-style (these must not be incorporated if you don't want to...)

delira/utils/config.py
gedoensmax: What do you prefer in that case ? Subclassing the type or just setting it to `bool` because this does not seem to work.

delira/utils/config.py
gedoensmax: I do not want to pass a class object i want to offer the possibility to just write a `type` into a `Config` and the `argparser` will expect this type then.

delira/utils/config.py
justusschock: Just using `isinstance(val, bool)` or even `isinstance(val, int)` (as `bool` is a subclass of `int`) for our check here

delira/utils/config.py
gedoensmax: I guess you mean `or` instead `of` right ?

justusschock: Right, that's a typo

delira/utils/config.py
justusschock: okay, in that case I'm still not sure, whether this will work, but at least this should be `issubclass` instead of `isinstance` because every object is an instance of `type` but you want to check for class-objects only...

delira/utils/config.py
gedoensmax: They will be retained as the old `Config` should be updated by the `argparser` output and not overwritten.

delira/utils/config.py
gedoensmax: Why would i only check for class objects and not types ?

delira/utils/config.py
gedoensmax: Ok so i will check against int and set the type to `type(val)` so that it is possible to set a boolean by giving a 0 or 1 ok ?

delira/utils/config.py
justusschock: Every type is a class object. In python everything is a object (including classes/types), but currently you check for all objects

delira/utils/config.py
gedoensmax: I am not sure how to do this as we can have multiple arguments in a `Config` with the same first letter and even the same objects within different `dicts` and so on.

delira/utils/config.py
gedoensmax: You are right!

delira/utils/config.py
gedoensmax: Ok i think i know what you mean i just can't think of an example for better unterstanding.

tests/utils/test_config.py
```diff
@@ -215,6 +218,48 @@ def test_internal_type(self):
...
+                    '--setflag']
+        # placeholder pyfile because argparser omits first argument from sys
+        # argv
+        with patch.object(sys, 'argv', ['pyfile.py'] + testargs):
+            cf.update_from_argparse(add_unknown_items=True)
+        self.assertEqual(cf['shallowNum'], int(testargs[1]))
+        self.assertEqual(cf['deep']['deepStr'], testargs[3])
+        self.assertEqual(cf['testlist'], testargs[5:7])
+        self.assertEqual(cf['setflag'], True)
+        with warnings.catch_warnings(record=True) as w:
```
justusschock, line 258: NIT: it would be better to use [assertWarns](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarns) for this, but this way it should also work.

justusschock (approved):
I minor comment regarding the code quality. Must not be done, just for completeness.

tests/utils/test_config.py
ORippler: thanks for the review, I would suggest merging as is

gedoensmax:
@justusschock Hey could we change this line https://github.com/delira-dev/delira/blob/e33bdb93135007748f0f2e155bf5d080ea4b9769/delira/utils/config.py#L547 ? You proposed to do it like it is but at least with `tf` it fails with some object or whatever in the dict i did not look too much into it but with `isinstance` instead of `issubclass` it works as intended.

PR #234, opened 2019-10-31
Fixed logging problems

### LINKED ISSUE #230: [Bug] Logging frequency image logging
We will have to look into TensorboardX logging again. If i turn up my logging frequency training goes well until then but stops at logging iteration with the error 
```
  File "/home/maxmueller/Documents/01_machine_learning/Paper_AE_ST_Samples/utils/logging.py", line 22, in at_iter_end
    'tag': 'Reconstructed_Image'}})
  File "/home/maxmueller/miniconda3/envs/paper_ae/lib/python3.7/site-packages/delira/logging/base_logger.py", line 269, in log
    super().log(log_message)
  File "/home/maxmueller/miniconda3/envs/paper_ae/lib/python3.7/site-packages/delira/logging/base_logger.py", line 206, in log
    raise e
  File "/home/maxmueller/miniconda3/envs/paper_ae/lib/python3.7/site-packages/delira/logging/base_logger.py", line 192, in log
    self._reduce_types[k])
  File "/home/maxmueller/miniconda3/envs/paper_ae/lib/python3.7/site-packages/delira/utils/dict_reductions.py", line 208, in reduce_dict
    if all([_v == v[0] for _v in v[1:]]):
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
the logging call looks like this with a resulting log message as shown below. 
Images have dimension (2,3,128,128) and are of type ndarray so everything as expected i would say.
Weird thing is that the `len(self._logging_queues[queue_key])` within `Logger.log` return `10` (logging frequency).
I would expect a list of only length one as it is the first logging that should be performed and all other loggings should just be dropped and not queued, right? Or do i have to handle this in my logging callback because if so i would actually like to change that and still it should not crash at this point.
```
self._logger.log(
            {'images': {'img_tensor': (data_dict['pred'][:trainer.module.n_log_images, :-1] + 1) / 2,
                        'tag': 'Reconstructed_Image'}})
```

NKPmedia:
A logging frequency of 2 or more will cause long waiting times: Fixed
Comparing images caused an error: Fixed

Closes #230

### DIFF
--- a/delira/logging/base_backend.py
+++ b/delira/logging/base_backend.py
@@ -151,7 +151,7 @@ class BaseBackend(object, metaclass=ABCMeta):
 
         """
         # get item from dict
-        process_item = self._queue.get(timeout=0.5)
+        process_item = self._queue.get(timeout=0.001)
         # log item if item is dict
         if isinstance(process_item, dict):
 
--- a/delira/utils/dict_reductions.py
+++ b/delira/utils/dict_reductions.py
@@ -205,7 +205,11 @@ def reduce_dict(items: list, reduce_fn) -> dict:
 
     for k, v in result_dict.items():
         # check if all items are equal
-        if all([_v == v[0] for _v in v[1:]]):
+        equals = [_v == v[0] for _v in v[1:]]
+        for idx, equality in enumerate(equals):
+            if isinstance(equality, np.ndarray):
+                equals[idx] = equality.all()
+        if all(equals):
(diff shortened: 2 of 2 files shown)


gedoensmax:
Have you checked if this also solves #230?

NKPmedia:
I think so. Justus and I fixed this together.

justusschock (approved):

PR #235, opened 2019-12-02
Add onecycle

NKPmedia:


ORippler (approved):
this is just a duplicate for travis pipe, see #233

PR #237, opened 2020-02-05
New license

cwern1:
As requested by our PI, we change the license to AGPL and add a copyright notice.


justusschock (approved):

justusschock:
I don't know, what's the issue with the torchscript backend. But since this PR does not affect it at all, go ahead and merge it

